PR #1, opened 2022-07-01
DOP-3097: Gatsby frontend boilerplate

drew-beckmen:
### Ticket

[DOP-3097](https://jira.mongodb.org/browse/DOP-3097): Basic Gatsby TS App
[DOP-3102](https://jira.mongodb.org/browse/DOP-3102): Jest and React Testing Library
[DOP-3099](https://jira.mongodb.org/browse/DOP-3099): Linting, Formatting, and Pre-Commit Hooks

### Notes

This PR includes the initial boilerplate code that will allow us to start developing the frontend of the application. In addition to setting up a basic Gatsby app with Typescript support, I've configured Jest and React Testing Library. I also included an example test for each of the libraries.

Lastly, this PR introduces pre-commit hooks for staged files when committing `js, ts, tsx, jsx` files from the `frontend` directory.


.github/workflows/frontend_test.yml
```diff
@@ -0,0 +1,26 @@
+name: Frontend CI
```
drew-beckmen, line 1: This may be replaced by a Drone pipeline, but I figured it would be helpful to ensure we maintain good test coverage and pass every test during early development.

frontend/README.md
```diff
@@ -0,0 +1,46 @@
...
+Once inside the `frontend` directory, run `npm run develop`. The site should now be running at `http://localhost:8000`.
+
+To build and serve the site, run:
+
+```shell
+$ npm run build
+$ npm run serve
+```
+
+Instructions on running via Docker are forthcoming.
```
rayangler, line 16: Part of the AC for DOP-3097 is to include a Dockerfile for the frontend. Is the plan to handle this separately now?

.github/workflows/frontend_test.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
rayangler, line 12: Just curious about your opinion here: Should these tests be runnable on MacOS as well, or are we better off sticking to only Ubuntu?

frontend/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
+.cache/
+public
+src/gatsby-types.d.ts
```
rayangler, line 4: Should we also include `.env.*` to ignore?

rayangler (commented):
Looking good, just a few questions for now

.github/workflows/frontend_test.yml
rkhullar-mongodb: i'll add that the ci pipeline only needs to test on linux, but yes locally it'd be ideally to work in docker or mac

frontend/.gitignore
rkhullar-mongodb: suggestion: i usually use [gitignore.io](https://www.toptal.com/developers/gitignore/) to generate this file

.github/workflows/frontend_test.yml
```diff
@@ -0,0 +1,26 @@
...
+    defaults:
+      run:
+        working-directory: ./frontend
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
```
rkhullar-mongodb, line 18: can you check if there are newer versions of these plugins?

frontend/README.md
rkhullar-mongodb: i'm fine with having that as another PR linked to the same ticket

.github/workflows/frontend_test.yml
drew-beckmen: I spent the last couple of hours getting Docker to work locally on my Mac with both `docker run`/`docker build` to launch the frontend server independently and `docker compose` so we can launch all the app's services from the project root. Will push these changes along with README updates shortly!

frontend/.gitignore
drew-beckmen: Cool tool! will leverage it 😄

frontend/README.md
drew-beckmen: added Docker to this PR

.github/workflows/frontend_test.yml
drew-beckmen: updated to v3. great catch!

.github/workflows/frontend_test.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        node-version: '16.x'
+    - name: Install dependencies
+      run: npm install --dev
```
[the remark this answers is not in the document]
drew-beckmen: relic of copy and paste, will remove 😓

frontend/Dockerfile.dev
```diff
@@ -0,0 +1,9 @@
+FROM node:alpine
+WORKDIR /app
+COPY ./package.json .
+RUN apk add --update python3 make g++\
```
rkhullar-mongodb, line 4: why do we need those packages?

rkhullar-mongodb: also, if we need to run the `apk add` we could use the `--no-cache` flag instead of removing the folder afterwards

drew-beckmen: The build failed without the Python dependency, and it seems the root cause is likely related to my [M1 arm architecture](https://github.com/docker/getting-started/issues/124)

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.4'
```
rkhullar-mongodb, line 1: nit: i suggest only specifying version `3`. the latest docker compose version is actually 3.8 but if we specify the minor version we're going to have to keep updating it, right? i don't think we'll have to worry about the minor version for docker itself

drew-beckmen: makes sense!

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.4'
+
+services:
+  hapleyfrontend:
+    image: hapleyfrontend
+    build:
+      context: frontend
+      dockerfile: ./Dockerfile.dev
```
rkhullar-mongodb, line 8: curious about the filename `./Dockerfile.dev`. why not just use `Dockerfile`?

drew-beckmen: I figured there would be additional considerations if we were to use a Dockerfile to build and host the frontend for production. For example, we only want to continuously run the Gatsby server in development. I'm not sure what the end state looks like in terms of building and deployment, but that was my logic...

frontend/Dockerfile.dev
drew-beckmen: it adds ~20s to an uncached image build, which isn't ideal, but not sure if there's a workaround

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  hapleyfrontend:
+    image: hapleyfrontend
+    build:
+      context: frontend
+      dockerfile: ./Dockerfile.dev
+    volumes:
+      - ./frontend/src:/app/src # Links the source files to the running container
+    ports:
+      - 8000:8000
```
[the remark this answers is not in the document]
drew-beckmen: switched over to port 3000. What do you think is the best approach in terms of environment variables as we'll need a `.env` at the root to use in the `docker compose` file, but we'll also need one within `frontend` so we can access `process.env`. Also I'm not sure there's a clean way to access environment variables within the `package.json` by pulling from `process.env`

frontend/Dockerfile.dev
rkhullar-mongodb: can you try installing rosetta (https://docs.docker.com/desktop/mac/apple-silicon/).
it's supposed to make it so the m1 can emulate `linux/amd64` images. i think you would just have to specify the platform in the docker compose file

docker-compose.yml
rkhullar-mongodb: adding here @drew-beckmen and I talked out of band about this.
my opinion on this is that the dockerfile should primarily be used for production. so if we want hot reloading we'd run  `npm run develop` on our local machines, but in docker we would not have that. in docker we'd just use commands for build and start / run.

there is a way to support a development mode with hot reloading in docker, but it'd make the dockerfile more complex. it can be done by using multiple stages and a build arg for the mode.

docker-compose.yml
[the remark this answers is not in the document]
rkhullar-mongodb: i think in the actual ts code we should use `process.env.SERVICE_PORT`. in the docker-compose we can either pass through environment variables from the host to the container, or we could just hardcode the port: i.e.:
```
    environment:
      - ENVIRONMENT=dev
       - SERVICE_PORT=3000
      - AWS_PROFILE=${AWS_PROFILE}
```

as for setting environment variables for local development, i like to use `direnv`. https://direnv.net/
it let's us configure the shell based on which directory we're in.

frontend/Dockerfile.dev
drew-beckmen: even with rosetta and specifying the architecture in the docker compose file, I get:
```
#0 482.9 npm ERR! code 1
#0 482.9 npm ERR! path /app/node_modules/@parcel/namer-default/node_modules/lmdb
#0 482.9 npm ERR! command failed
#0 482.9 npm ERR! command sh -c node-gyp-build
#0 482.9 npm ERR! gyp info it worked if it ends with ok
#0 482.9 npm ERR! gyp info using node-gyp@9.0.0
#0 482.9 npm ERR! gyp info using node@18.4.0 | linux | x64
#0 482.9 npm ERR! gyp ERR! find Python 
```

docker-compose.yml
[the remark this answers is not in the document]
drew-beckmen: The port is passed to the `gatsby develop` or `gatsby serve` command via scripts in `package.json`. For example,
```
  "scripts": {
    "develop": "gatsby develop -p 3000",
```

Gatsby provides support for loading of environment variables into the browser and (optionally) Node.js. I'd like to leverage Gatsby's built-in support for env variables since they account for environment specific `.env` files (`.env.development`, `.env.production`). However, I don't see a way to pull environment variables from `.env.development` into `package.json` unless we load them into the shell (bypassing Gatsby's recommended handling of env vars)https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/

docker-compose.yml
drew-beckmen: I updated the Dockerfile to reflect production environment. I tried tons of different setups to avoid needing the Python dependency, but nothing else would work on my M1 so I'm going to leave it for now. The initial image build takes a while, but subsequent cached builds are quick.

frontend/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:alpine
```
rkhullar-mongodb, line 1: nit: can you specify the node version here?

frontend/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:alpine
+WORKDIR /app
+COPY ./package.json .
+RUN apk add --no-cache --update python3 make g++
+RUN npm install
+COPY . .
```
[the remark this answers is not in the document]
drew-beckmen: pulling in `src` directory along with TS and Gatsby config files, which is sufficient to build the application

PR #2, opened 2022-07-05
DOP-3103: Add email addresses to entitlements collection

drew-beckmen:
### Ticket

[DOP-3103](https://jira.mongodb.org/browse/DOP-3103)

### Notes

Python script that leverages Realm endpoint to add email addresses to each of the users on the `pool.entitlements` collection when run with `--prod` flag. By default, runs with `--dev` flag (changes `pool_test.drew_entitlements_dev`). Can also be run with `--test`, which writes to `pool_test.entitlements`. 

Also allows users to run with a `--rollback` flag to drop email addresses if a rollback becomes necessary. 

**Known Issue**:
The Realm endpoint returns a 404 for interns and inactive users. I'll manually add emails where necessary (should be about 10).


scripts/add_emails.py
```diff
@@ -0,0 +1,91 @@
+"""
+    DOP-3103: Add email address to every user
+    Author: Drew Beckmen
+
+    The goal of this ticket is to add a populated email address
+    field to each document in the entitlements collection in the pool database.
+    We use a Realm endpoint from Mana to find the email address for each user.
```
rayangler, line 7: (Nit) I would add a usage instruction here. Maybe reference that `.env.example` should be used for this script to make the relationship clear for future context?

rayangler (approved):
LGTM, but I'm curious if we should be keeping the `add_emails.py` script in this repo in the first place. I was under the impression that this wouldn't be a script we would run frequently (since a user's email address can be obtained when we add a new user in the future)

drew-beckmen:
> LGTM, but I'm curious if we should be keeping the `add_emails.py` script in this repo in the first place. I was under the impression that this wouldn't be a script we would run frequently (since a user's email address can be obtained when we add a new user in the future)

Agreed that the script does not need to live in the repository long term. I originally put it in my own GitHub repo just to better facilitate reviews, but this approach seemed cleaner. Now that we have the approval, I'm happy to either:

1. merge it in, run the script, and open a follow-up PR to delete the `scripts` directory
2. close the PR and run it locally

Either work for me. I think opening the PR just helped facilitate getting 👀 , and it also leaves a bit more of a paper trail when looking at the history of the project.

rayangler:
Ah, okay, thanks for the context on that! I think it makes sense to just close the PR if we're going to delete the scripts directory after merging anyways. It would save us a step

PR #3, opened 2022-07-05
DOP-3098: Set up boilerplate for FastAPI backend

rayangler:
### Links

[DOP-3098](https://jira.mongodb.org/browse/DOP-3098)

### Notes

* Adds basic dependencies and files for the backend.
* I'm having a tough time getting Docker working on my computer right now. I might have to make it a separate ticket if I can't get it working soon.
* Let me know if I'm missing something else that we should have or set up.


backend/.gitignore
```diff
@@ -0,0 +1,6 @@
+__pycache__/
+.env
+.venv
+
+# pipenv
+Pipfile.lock
```
rayangler, line 6: It seems like there are different opinions regarding whether `Pipfile.lock` should be ignored or not. I figured ignoring it for now and having the package versions in the `Pipfile` will still get the job done, but please correct me if this is the wrong assumption

backend/README.md
```diff
@@ -0,0 +1,42 @@
+# Hapley Backend API
+
+## Local Development
+
+### Requirements and Installation
```
drew-beckmen, line 5: maybe add a note in the README for how to access the auto-generated API documentation?

backend/README.md
```diff
@@ -0,0 +1,42 @@
...
+python3 -m pipenv install
+```
+
+### Running Locally
+
+```
+uvicorn main:app --reload
+```
+
+Go to `http://127.0.0.1:8000/` to access routes.
```
drew-beckmen, line 30: Gatsby also runs on port `8000` by default. Do you think we could run the backend on `8001`? Otherwise we'll have a conflict when developing both services simultaneously.

drew-beckmen (commented):
Overall LGTM, only blocking concern is that we need to decide on ports for the frontend/backend since both run in `8000` by default

backend/.gitignore
rkhullar-mongodb: also, this is something we could don in another PR, but shouldn't we only have one gitignore file? and also keep the ignore file and precommit hooks under `.github` under the project root?

backend/Pipfile
```diff
@@ -0,0 +1,14 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+fastapi = "~=0.78.0"
+uvicorn = "~=0.18.2"
+pydantic = "~=1.9.1"
```
rkhullar-mongodb, line 9: nit: when we use `~=` the patch version will get ignored and actually default the latest.
i.e: `~=0.18.0` == `~=0.18.2` == `~=0.18.X`
so i like to just leave the patch version as `0`

backend/.gitignore
rayangler: > so i think it's fine for us to commit the lock file since we're all on mac. it might become an issue if we come across a library that depends on a compiled extension, since we don't all have the m1.

Yup, the comment from gitignore.io was actually what swayed me into ignoring the file. The 3 of us may all be on Macs but I believe Andrew uses Linux. I'm okay with removing it from the `.gitignore` for now and then revisiting it in the future if it actually proves to be an issue though. I'll go ahead and push up the `Pipfile.lock`.

> also, this is something we could don in another PR, but shouldn't we only have one gitignore file? and also keep the ignore file and precommit hooks under `.github` under the project root?

Yeah, we can always consolidate later. For what it's worth, I don't think having more than one `.gitignore` is necessarily bad and is supported (I think the file paths listed in a `.gitignore` will just be relative to the directory it's in)

drew-beckmen: I've moved the pre-commit hook script into the project root under `.husky` in my frontend PR (based on instructions in this [blog post](https://dev.to/monfernape/enforce-husky-pre-commit-with-eslint-prettier-in-monorepo-55jc)). This _should_ allow Raymund to hook into the Git commit lifecycle as well by adding to the `.husky/pre-commit` script

backend/README.md
rayangler: Added a section for it after running locally!

backend/main.py
```diff
@@ -0,0 +1,7 @@
+from fastapi import FastAPI
+
+from routers import hello_world
```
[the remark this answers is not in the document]
rayangler: Added the files!

rkhullar-mongodb:
This looks like a good start to me!

rayangler:
@rkhullar-mongodb Added Docker-related stuff to the PR. Might need a follow-up to consolidate anything between the frontend and backend but curious if you have any thoughts about what's been pushed up

backend/.dockerignore
```diff
@@ -0,0 +1,5 @@
+__pycache__/
+.env
+.venv
+Dockerfile
```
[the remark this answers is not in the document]
rayangler: Yup, I can go ahead and move any specific ignores (I assume something like `__pycache__`) to that root `.dockerignore`.

backend/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.10-bullseye
```
rkhullar-mongodb, line 1: curious about `bullseye`. is that a stable debian version? and why not use `slim` or `alpine`?

backend/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+WORKDIR /app
+COPY ["Pipfile", "Pipfile.lock", "./"]
+RUN pip install pipenv \
+    && pipenv install --system
+COPY . .
```
rkhullar-mongodb, line 6: i suggest copying the specific files and folders that we need in the docker image here, since we probably don't need the readme or other future files likes a potential `docs` folder or the drone / github action pipelines.

backend/Dockerfile
rayangler: `bullseye` seemed to be the stable version based off of [these packages](https://packages.debian.org/search?keywords=python3). I wasn't 100% sure which version to used so I selected the topmost simple tag in [this list](https://hub.docker.com/_/python) that wasn't a 3.11 beta/release candidate version.

Is there a preference for what we use?

backend/Dockerfile
rayangler: My understanding was that the `.dockerignore` file would prevent such files from being copied. Is this not the case? Looking at the terminal of the running Docker container, it doesn't seem like the ignored files (including the `README`) were copied. Just wondering for best practice

rkhullar-mongodb: the `.dockerignore` file does prevent files being copied over, but from my understanding the main reason to use that file is to limit the build context. it's to avoid sending large or many files to docker during the build like the `__pycache__` or `venv`.

we want to optimize the docker image and only keep useful files inside of it. IMO it's simpler to have an "allow" list and specify those files in the dockerfile itself, than keeping an exhaustive "deny" list in the `dockerignore`

backend/.dockerignore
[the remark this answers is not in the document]
drew-beckmen: Edit: the `.dockerignore` needs to be in the same directory as the `Dockerfile` so it gets picked up in the build context. Ignore what I said before — we'll just use per-directory `.dockerignore` files

backend/Dockerfile
rayangler: Got it, thanks for this explanation! I've updated the `Dockerfile` accordingly. Let me know what you think

rayangler:
Dockerfile optimizations are still in question, but I'll go ahead and merge this to unblock initial Kanopy setup. We can optimize Docker in a follow-up if anything comes up, but running it locally works. Thanks all!

PR #5, opened 2022-07-08
DOP-3115: Set up Drone pipelines for staging builds and deployments via Kanopy on single domain

drew-beckmen:
## Ticket

[DOP-3115](https://jira.mongodb.org/browse/DOP-3115)
[DOP-3120](https://jira.mongodb.org/browse/DOP-3120)

### Drone Pipelines
This PR introduces a set of Drone pipelines. The first set of pipelines, triggered by a push to the `main` branch, run the test suites for the application and upload the Docker images to ECR via the Kaniko plugin. The frontend and backend tests run in parallel, and the uploads to ECR run only after all tests pass. Here's a graphical view of the pipeline stages from Drone:
Screen Shot 2022-07-08 at 2 47 14 PM

You can access a sample build [here](https://drone.corp.mongodb.com/mongodb/docs-hapley/7).

After uploading to ECR, the final step in the pipeline is to promote the build to staging. The promotion to staging kicks off another pipeline responsible for deploying the frontend and backend via the [web-app Helm Chart](https://github.com/10gen/helm-charts/tree/master/charts/web-app).

You can access a sample run of the deployment pipeline [here](https://drone.corp.mongodb.com/mongodb/docs-hapley/8).

**Testing Story**

I tested the pipelines by including a `push` trigger for this feature branch. I've since removed that trigger to ensure we only run the pipelines on pushes to the `main` branch.

### Staging Deployments, User Authentication

By default, all services on the Kanopy platform pass through the CorpSecure authentication proxy, which is the default authentication service for internal applications. As a result, we don't need to implement any custom authentication logic. Users that navigate to Hapley will be forced to login with Okta if they don't have an active session.
 
**Staging Links**

The Hapley frontend is now hosted at: https://hapley.docs.staging.corp.mongodb.com/
The Hapley backend is now hosted at: https://hapley.docs.staging.corp.mongodb.com/api/v1

By consolidating both services onto a single domain, we make it much simpler to send requests from the client to the backend server. CorpSecure, which is the authentication proxy put in front of all Kanopy services, only supports simple CORS requests (requests that don't cause side effects on server data). With both services on the same site, we won't have to deal with any frustrating CORS errors.

If you navigate to https://hapley.docs.staging.corp.mongodb.com/request-poc, you'll see a button that, when clicked, sends a request to the backend server through the simplest possible invocation of the `fetch` API.

### Pull Request Testing

We will rely on GitHub Actions for running our test suite on pull requests.

We plan to continue using GitHub Actions for PRs because of how easy it is to conditionally run tests based on changed files. The `setup-node` action also presents an elegant solution for caching project dependencies. This PR introduces dependency caching and conditional tests using the `dorny/paths-filter` [action](https://github.com/dorny/paths-filter). In the Action details, you can see that the linting & testing steps were skipped:

Screen Shot 2022-07-08 at 4 05 27 PM

Any merge into `main` will trigger both backend and frontend tests as described in the overview of the Drone pipelines above.

### Open Questions:

For the Drone pipelines:
- do we want to investigate building a pipeline for preview links on an individual PR? The devcenter team currently achieves this via Vercel, but they seem to be transitioning over to using Kanopy. POC PR is [here](https://github.com/mongodb/devcenter/pull/229/files).
- caching: how can we cache our Docker images across Drone builds? right now, the image is rebuilt each time, which is time consuming
- caching: how can we cache our dependencies Drone builds? running `npm install` takes too long 😢


frontend/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM node:16-alpine
+ARG GATSBY_API_URL=$GATSBY_API_URL
```
rayangler, line 2: Will there be a difference between the value of this env between staging and production? Or are they expected to be the same `localhost`?

rayangler (commented):
Mostly just questions for now!

.drone.yml
```diff
@@ -0,0 +1,156 @@
...
+      GATSBY_API_URL: https://hapley.docs.staging.corp.mongodb.com/api/v1
+    settings:
+      create_repository: true
+      enable_cache: true
+      dockerfile: ./frontend/Dockerfile
+      context: ./frontend
+      registry: 795250896452.dkr.ecr.us-east-1.amazonaws.com
+      repo: docs/${DRONE_REPO_NAME}/frontend
+      build_args:
+        - GATSBY_API_URL
```
rkhullar-mongodb, line 67: is this needed at build time? or can this be an environment variable on runtime for the frontend?

.drone.yml
```diff
@@ -0,0 +1,156 @@
...
+    - npm run test
+    when:
+      event:
+        - push
+
+trigger:
+  branch:
+    - main
+  event:
+    - push
```
rkhullar-mongodb, line 46: does the pipeline run on pull requests to main?

backend/tests/base.py
```diff
@@ -0,0 +1,8 @@
+from fastapi.testclient import TestClient
+from main import app
+
+class FastApiTest():
+    def __init__(self):
+        self.client = TestClient(app)
+        self.client.base_url += "/api/v1"
+        self.client.base_url = self.client.base_url.rstrip("/") + "/"
```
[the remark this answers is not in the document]
drew-beckmen: I'll remove this line and amend the previous line to be `self.client.base_url += "/api/v1/"`. Line 8 was meant to be a safety check based on the info in [this SO post](https://stackoverflow.com/questions/69166262/fastapi-adding-route-prefix-to-testclient), but I think it's ok to remove it.

backend/tests/routers/test_hello_world.py
```diff
@@ -1,11 +1,9 @@
...
+from ..base import FastApiTest
 
-from main import app
-
-client = TestClient(app)
+client = FastApiTest().client
 
 def test_hello_world():
-  response = client.get("/hello")
+  response = client.get("hello")
```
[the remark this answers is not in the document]
drew-beckmen: Under the hood, `TestClient` uses the `urljoin` method which requires `base_url` to end in a trailing slash and the part you are appending to **NOT** have a leading slash. Unfortunately, it doesn't seem there's an easy way around this unless we wanted to specify `/api/v1/foo` in every test.
https://stackoverflow.com/questions/69166262/fastapi-adding-route-prefix-to-testclient

.drone.yml
drew-beckmen: No it does not. We're using GitHub actions for pull request checks right now. Upon merging of a PR into main, this pipeline gets kicked off to promote the new build to staging.

.drone.yml
```diff
@@ -0,0 +1,156 @@
...
+      enable_cache: true
+      dockerfile: ./frontend/Dockerfile
+      context: ./frontend
+      registry: 795250896452.dkr.ecr.us-east-1.amazonaws.com
+      repo: docs/${DRONE_REPO_NAME}/frontend
+      build_args:
+        - GATSBY_API_URL
+      tags:
+      - git-${DRONE_COMMIT_SHA:0:7}
+      - latest
```
rkhullar-mongodb, line 70: we might do this in another PR, but I suggest different tags:
    - ${DRONE_SOURCE_BRANCH//\//-}
    - ${DRONE_SOURCE_BRANCH//\//-}-${DRONE_COMMIT_SHA:0:7}
    - drone-build-${DRONE_BUILD_NUMBER}
    
I don't think we should use "git-" as a tag prefix since that might make us think the image was built with github actions instead of drone.
 
 And I like to avoid publishing `latest` since that makes checking the running app version less clear later on.
 Instead, if we publish a tag for the `{branch-name}` , that'll help with basic docker build caching.

rkhullar-mongodb (approved):
Thanks for adding this @drew-beckmen !

As for the open questions I did add a comment about using branch names in tags to enable caching in the build pipeline. Preview environments would also be cool. I think those are good fast follows to this PR.

.drone.yml
drew-beckmen: Will amend these tags and the image value for the deployment pipeline!

backend/tests/base.py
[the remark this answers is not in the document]
rayangler: Oh okay, thank you for the context on that. Happy to keep as-is if that's the case. It might be nice to have a short comment to say that `TestClient` requires a single trailing slash for url joins for posterity

backend/tests/routers/test_hello_world.py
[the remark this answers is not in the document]
rayangler: Thanks for the explanation! I think the comment about `TestClient` and `base_url` that I mentioned in the other thread would be nice to have to help explain why this difference is necessary

backend/main.py
```diff
@@ -1,7 +1,18 @@
 from fastapi import FastAPI
+from starlette.middleware.cors import CORSMiddleware
 
 from routers import hello_world
 
 app = FastAPI()
 
+# Prevent CORS errors in local development
+origins = ["http://localhost:3000"]
```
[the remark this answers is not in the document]
rayangler: Makes sense, thanks!

.drone.yml
drew-beckmen: https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/

Based on my understanding, Gatsby requires it to be present at build time, otherwise the compiled Javascript will have `undefined` where `GATSBY_API_URL` is referenced

frontend/Dockerfile
drew-beckmen: yes! it is passed through as a build arg to the Dockerfile, so we can control the value in `.drone.yml` when we build the Docker image. When we add production pipelines down the road, we'll need to ensure that the environment variable is properly set to the production API URL.

backend/tests/base.py
```diff
@@ -0,0 +1,18 @@
...
+        self.base_url += "/api/v1/"
+
+    # TestClient uses urljoin, which requires a trailing slash on base_url
+    # and no leading slash on the second arg. https://stackoverflow.com/questions/69166262/fastapi-adding-route-prefix-to-testclient
+    def trim_leading_slash(self, path):
+        if path.startswith("/"):
+            path = path[1:]
+        return path
+
+    def get(self, path: str, **kwargs):
```
drew-beckmen, line 17: @rayangler does this look better? I tried to abstract away the logic from individual test files and handle all the `/` logic here. Now our tests can fetch `/foo` or `foo`, either will work

drew-beckmen:
> Thanks for adding this @drew-beckmen !
> 
> As for the open questions I did add a comment about using branch names in tags to enable caching in the build pipeline. Preview environments would also be cool. I think those are good fast follows to this PR.

For now, I think we'll hold on preview environments. It looks like dev center is working on it themselves right now, and there's some additional complexity given the limited number of actions/triggers Drone is able to respond to. Here's the PR for future reference: https://github.com/mongodb/devcenter/pull/229

I agree that caching in the build pipeline is a priority, and I'll create a follow-on ticket to work on caching.

backend/tests/base.py
rayangler: Interesting, I like the flexibility of this and it can potentially help with keeping things consistent across future tests. Needing to override the method isn't ideal, but let's keep what you have for now since it works for our current use case.

rayangler (approved):
LGTM again! Thanks for all of the contexts and for working on this!

PR #6, opened 2022-07-11
DOP-3101: Add testing to backend

rayangler:
### Ticket
DOP-3101

### Notes
- Adds testing with coverage to the backend.
- Adds `pipenv` script commands so that we don't have to remember as many commands to run the app and to run tests.
- Adds Github actions job for backend tests. Will require consolidation/resolving of any merge conflicts with frontend PR.
- `requests` has been added as a dev dependency since it's apparently used for testing but not required for production.


.github/workflows/pull-request-checks.yml
```diff
@@ -0,0 +1,77 @@
...
+        uses: actions/checkout@v3
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          cache: 'npm'
+          cache-dependency-path: '**/package-lock.json'
+      - name: Install frontend dependencies
+        run: npm ci
+      - uses: dorny/paths-filter@v2
```
drew-beckmen, line 66: we can remove this now that you've added the path filtering as a separate check

rayangler: Done! Thanks for catching that

PR #7, opened 2022-07-12
DOP-3100: Add formatting and linting to the backend

rayangler:
### Ticket

DOP-3100

### Notes

* Uses `pre-commit` to run pre-commit hooks. See [this section](https://github.com/mongodb/docs-hapley/blob/8991c4638ac50c4ff5bb010d705697520a6b4ad3/README.md#pre-commit-hooks) on pre-commit hooks in the `README`. If you already had `husky` installed locally, you may need to run `git config --unset-all core.hooksPath` in your local repo before running `pre-commit install`.
* Removes `husky` as a dependency for the frontend. `pre-commit` already does the job of running pre-commit hooks.

### DIFF
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -37,12 +37,14 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10.x'
-      - name: Install pipenv
+      - name: Install pipenv and dependencies
         run: |
           pip install pipenv
+          pipenv install --dev
+      - name: Format and lint backend
+        run: pipenv run format && pipenv run lint
       - name: Run tests
         run: |
-          pipenv install --dev
           pipenv run test
 
   frontend:
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           pip install pipenv
           pipenv install --dev
-      - name: Format and lint backend
+      - name: Format and lint
         run: pipenv run format && pipenv run lint
       - name: Run tests
         run: |
@@ -63,9 +63,9 @@ jobs:
           node-version: '16.x'
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
-      - name: Install frontend dependencies
+      - name: Install dependencies
         run: npm ci
-      - name: Lint Frontend
-        run: npm run lint && npm run format
-      - name: Test Frontend
+      - name: Format and lint
+        run: npm run format && npm run lint
+      - name: Run tests
         run: npm test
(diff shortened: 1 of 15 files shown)


.github/workflows/pull-request-checks.yml
```diff
@@ -37,12 +37,14 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10.x'
-      - name: Install pipenv
+      - name: Install pipenv and dependencies
         run: |
           pip install pipenv
+          pipenv install --dev
+      - name: Format and lint backend
```
drew-beckmen, line 44: very nit but we can probably remove references to "backend" within steps in the backend job (the same holds true for the frontend job). It just seems redundant to say "lint backend" when the GH Action interface will already show the job name as backend

README.md
```diff
@@ -16,6 +16,29 @@ To read more about the application backend, see the [README](./backend/README.md
 
 ## Development and Deployment Workflow
 
+### Pre-commit Hooks
+
+This monorepo uses [pre-commit](https://github.com/pre-commit/pre-commit) to 
+help lint and format both backend and frontend files before they are committed 
+through git.
+
+With python 3.10, install and set up `pre-commit` for this repo by doing the
```
drew-beckmen, line 25: thanks for diving into `pre-commit`!

backend/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore = E501
```
drew-beckmen, line 2: just out of curiosity, what is this file?

README.md
```diff
@@ -16,6 +16,29 @@ To read more about the application backend, see the [README](./backend/README.md
...
+
+With python 3.10, install and set up `pre-commit` for this repo by doing the 
+following:
+
+```
+pip install pre-commit
+pre-commit install
+```
+
+This will allow `pre-commit` to be run when doing `git commit`. To test pre-commit
```
drew-beckmen, line 33: could you add a brief description of the linting and formatting tools we use in the backend `README`? Perhaps just add a section called Linting below Testing?

drew-beckmen (approved):
LGTM! I'd love to see a small addition to the backend README with a recap of the tools we use (black and flake8?), but no blocking concerns

👏  for 🧽 🧼 code

.github/workflows/pull-request-checks.yml
rayangler: Good suggestion! Done

README.md
rayangler: Hopefully less susceptible to weirdness with a monorepo than `husky` was for us 🤞

backend/setup.cfg
rayangler: It's a general setup file for different packages. I think it's also similar to `pyproject.toml`, and `flake8` has a dedicated version of this called `.flake8`. I wasn't sure what the preferred file was but since `flake8` mentioned `setup.cfg` in their [docs](https://flake8.pycqa.org/en/latest/user/configuration.html?highlight=setup.cfg#project-configuration), I went with that. 😅 

As for the contents of this file currently, it establishes the configuration and rules of linting for `flake8`. Just added a comment to make what `E501` represents explicit. (I think it's sometimes unavoidable to have long lines due to stuff like links in comments)

PR #8, opened 2022-07-12
DOP-3126: Custom middleware for user authorization based on Okta groups, connection to MongoDB with Beanie

drew-beckmen:
### Ticket

[DOP-3126](https://jira.mongodb.org/browse/DOP-3126)
[DOP-3125](https://jira.mongodb.org/browse/DOP-3125)

#### Middleware
Although Kanopy provides authentication for us out of the box via the CorpSecure authentication proxy, we don't necessarily want our API endpoints to be accessible to anyone within MongoDB. Without authorization, any MongoDB employee could introduce server-side data changes by passing an Okta `auth_token` with their request.

To solve this issue, we are introducing user authorization based on Okta groups. 

**Staging and Production**
In `staging` and `production` environments, Kanopy automatically attached the JWT token to the `Authorization` headers of the request. This PR adds custom middleware that validates the JWT token belongs to a user who is a member of an Okta group with access. For now, I've limited access to `10gen-docs-platform`. This will suffice during our early development stages as no docs content writers will need access yet. Once we are reading to do UAT, we can modify the authorized groups.

**Local Development**
The backend README has been updated with new instructions on how to generate a fake JWT token to use during local development. Developers will need to request a token from the `/api/v1/login` endpoint, which they can then put in the `.env` file. If `Authorization` headers are not present on the request, the middleware will read from the `JWT_TOKEN` environment variable.

The middleware also attaches information extracted from the JWT token to each individual request. This will provide us with access to the Okta username and email of the user making the request.

With this approach, we get:
- no conditional branching of code based on environment
- the ability to enrich requests with user info in local development based on the decoded fake JWT

**401 Forbidden Status Codes**

The backend will return a 401 error under the following circumstances:
- the token is missing or invalid (raises a `JWTError`)
- the token is present and valid, but associated with a user who isn't a member of the correct Okta groups (raises a `HTTPException`)

#### Refactoring
This PR also introduces some broader refactors inspired by the project structure of [Rajan's FastAPI POC ](https://github.com/rkhullar/hello-fastapi). I think this project structure helps split up code into smaller, more digestible files. It also provides a natural pattern to prefix our API routes with `/api/v1`, which is expected after #4. Rather than manually adding `prefix="/api/v1"` to the router, this project structure organizes files in a way where we could theoretically create a `v2` quite easily without significant refactoring.

#### Database Connection

This PR introduces the Beanie ODM as our way of interacting with data in MongoDB. I've also added instructions to the README for easily setting up the database for local development with seed data. I implemented a single endpoint, `GET /repos`, which returns a list of all the repositories to which a user is entitled. The user information gets attached to the request via middleware after the JWT token is decoded.

**Open Question**:
- In a follow-up, I'll experiment with different methods of testing endpoints that interact with the database.


rayangler:
> I don't love the environment-specific middleware. Do others have better ideas regarding implementation? In general, I don't love writing conditional logic on environment variables.

Would it make sense to just include both middlewares on development and only the `Authorization` middleware in staging/production, instead of choosing only 1 over the other? The CORS middleware seems unavoidable for local development regardless.

> How will we identify users' Okta IDs when developing locally? Should we just have them hard-code it in a .env file? This also creates environment specific logic because in staging/prod, our requests would automatically get enriched with user data, while in development/test they would not.

We can create a JWT token for a fake user with other data we need, such as groups, email, etc., locally to pass into the API requests when developing/testing. If accessing through the frontend, we can pass along that JWT token in our requests to the backend.

drew-beckmen:
> Would it make sense to just include both middlewares on development and only the `Authorization` middleware in staging/production, instead of choosing only 1 over the other? The CORS middleware seems unavoidable for local development regardless.

I'm fine with this change. I agree that it simplifies the logic.

> We can create a JWT token for a fake user with other data we need, such as groups, email, etc., locally to pass into the API requests when developing/testing (unless we want to use our actual `auth_token`s for local development). If accessing through the frontend, we can pass along that JWT token in our requests to the backend.

This sounds good, except I wonder how it will look on the frontend because in staging/production, CorpSecure automatically passes the `auth_token` with each request, so we don't need to specify any `headers` when calling `fetch` (in fact, we couldn't even if we wanted to since those cookies are `httponly`). I wouldn't want to have to include request headers with a fake JWT token in development and then remove those headers prior to merging.

rayangler:
> This sounds good, except I wonder how it will look on the frontend because in staging/production, CorpSecure automatically passes the auth_token with each request, so we don't need to specify any headers when calling fetch (in fact, we couldn't even if we wanted to since those cookies are httponly). I wouldn't want to have to include request headers with a fake JWT token in development and then remove those headers prior to merging.

Ah, okay! We can cross the exact/actual implementation details of the frontend locally when we get there, but in that case we'd probably have to rely on an env variable that can be used to conditionally add the authorization header in development.

backend/api/core/middleware/authorization.py
```diff
@@ -0,0 +1,70 @@
...
+            if bool(self.AUTHORIZED_OKTA_GROUPS & set(token_data.groups)):
+                request.state.user = token_data
+                response = await call_next(request)
+            else:
+                raise UnauthorizedOktaGroup()
+        except (UnauthorizedOktaGroup, JWTError) as e:
+            response = JSONResponse(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                content={
+                    "message": f"Unauthorized. If developing locally, see the README for instructions on creating a JWT token. Error message: {e}"
```
rayangler, line 49: Is there a difference between `UnauthorizedOktaGroup` and `JWTError` such that we can handle their errors separately? It feels a little weird to show "If developing locally" in production to users who are intended to not have authorization.

backend/api/v1/routes/login.py
```diff
@@ -0,0 +1,9 @@
+from fastapi import APIRouter
+from ...core.middleware.authorization import Authorization
+
+router = APIRouter()
+
+
+@router.get("/login")
+async def login(email: str, username: str):
```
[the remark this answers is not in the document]
drew-beckmen: I agree that the naming could be improved. I'll update it to `/sample-token` to illustrate that this endpoint will not be used to login or generate any real/meaningful tokens.

backend/api/core/middleware/authorization.py
drew-beckmen: Good call. We'd only ever get `UnauthorizedOktaGroup` when a MongoDB user without valid Okta groups tries to access a resource, so we shouldn't need to provide hints about local development in that situation.

backend/api/core/config.py
```diff
@@ -0,0 +1,20 @@
...
+    description: str = "API for Hapley, the MongoDB docs platform management tool"
+    mongo_uri: str = getenv("MONGO_URI")
+    mongo_db_name: str = getenv("MONGO_DB_NAME")
+
+    class Config:
+        env_file = ".env"
+        env_file_encoding = "utf-8"
+
+
+@lru_cache()
```
rkhullar-mongodb, line 18: nit: why do we need to cache the settings?
since the settings are instantiated and include in the `app` object, aren't we only reading the settings on startup anyways?

backend/seed/entitlements.json
```diff
@@ -0,0 +1,381 @@
...
+    "10gen/cloud-docs-osb",
+    "mongodb/docs-compass",
+    "10gen/docs-kafka-connector",
+    "mongodb/docs",
+    "10gen/cloud-docs",
+    "mongodb/docs-compass",
+    "10gen/cloudgov-docs"
+  ],
+  "email": "matt.meigs@mongodb.com"
+}]
```
rkhullar-mongodb, line 381: i'm not sure about including this file, since it does have user information like our email, github, and slack

backend/README.md
```diff
@@ -28,11 +28,35 @@ The `--dev` option allows both production and dev packages to be installed.
...
+4. Create a `JWT_TOKEN` environment variable in `.env`. Set this equal to the token received in the JSON response.
+5. Shut down the development server with Ctrl + C
+
+**Seeding the Database**:
+
+Before you begin these steps, ensure that you've installed the MongoDB database and shell on your computer. You can find instructions for installing MongoDB [here](https://www.mongodb.com/docs/manual/installation/) and instructions for installing `mongosh` [here](https://www.mongodb.com/docs/mongodb-shell/install/).
+
+After installing MongoDB, don't forget to run it to start the `mongod` process. The command for running on macOS would be `brew services start mongodb-community@<version #>`.
+
+To verify that the installations were successful, run the command `mongosh`. The Mongo shell should connect to your local `mongod` process automatically.
```
rkhullar-mongodb, line 46: just adding this here as a note: i think there's a couple other options for accessing the database for development.

for purely local runs, we either have mongo running directly on our machines, or we leverage docker-compose. 
for development where we need actual data, maybe it would make sense to have a shared `dev` database in atlas.

backend/seed/entitlements.json
drew-beckmen: good point i always forget all our repos are public here 🤦 i'll fix it and update instructions for loading seed data based on an export from Mongo

backend/README.md
drew-beckmen: I'd be curious to hear if @rayangler has a preference for developing with an Atlas database or with `mongod` running locally

rayangler: I personally don't mind either, but I might lean a little towards a local instance. I think the main benefit of having a local instance running is that we can edit the data however we want. With an Atlas database, there's the risk of accidentally changing data that someone else needs. Locally, we can always mess up everything however we'd like and then easily load everything back to their initial state as needed.

rayangler: That being said, there's nothing stopping us from doing both and having a shared dev db as well. Developers will just choose which one they want to connect to

backend/README.md
```diff
@@ -28,6 +28,17 @@ The `--dev` option allows both production and dev packages to be installed.
 
 ### Running Locally
 
+**Generating a JWT Token**:
+The API implements middleware that validates incoming JWT tokens are associated with users who belong to authorized Okta groups. Follow these steps to generate a fake JWT token that can be used for local development:
+
+1. Create a `.env` file within the backend directory.
+2. Start the local development server with `pipenv run app`
+3. Using Postman, `cURL`, or a similar tool, send a request to the `/api/v1/sample-token` endpoint with `username` and `email` query parameters. For example: `http://localhost:8000/api/v1/sample_token?email=foo@mongodb.com&username=foo`
```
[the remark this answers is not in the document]
drew-beckmen: yes good catch!

backend/README.md
```diff
@@ -28,11 +28,35 @@ The `--dev` option allows both production and dev packages to be installed.
...
+4. Create a `JWT_TOKEN` environment variable in `.env`. Set this equal to the token received in the JSON response.
+5. Shut down the development server with Ctrl + C
+
+**Seeding the Database**:
+
+Before you begin these steps, ensure that you've installed the MongoDB database and shell on your computer. You can find instructions for installing MongoDB [here](https://www.mongodb.com/docs/manual/installation/) and instructions for installing `mongosh` [here](https://www.mongodb.com/docs/mongodb-shell/install/).
```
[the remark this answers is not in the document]
drew-beckmen: will do!

backend/README.md
drew-beckmen: Ok! I'll add options for both into the README!

environments/backend-staging.yaml
```diff
@@ -1,6 +1,12 @@
 service:
   targetPort: 8000
 
+env:
+  MONGO_DB_NAME: pool_test
+
+envSecrets:
+  MONGO_URI: docs-hapley
```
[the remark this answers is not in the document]
rkhullar-mongodb: docs-hapley is the name of the secret in k8s,
and then the secret has key-value pairs.
Screen Shot 2022-07-15 at 3 47 26 PM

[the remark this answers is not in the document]
rkhullar-mongodb: https://kanopy.staging.corp.mongodb.com/docs/secrets/docs-hapley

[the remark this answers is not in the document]
rayangler: Oh, thank you!

backend/api/core/config.py
drew-beckmen: In `root.py`, I use the `get_settings()` function to return information stored about the application in the JSON response. In the test files, I've changed any calls to `get_settings()` to reference the settings on the `app` associated with the `TestClient`, but I'm not sure how I would be able to reference `app.settings` in `root.py`.

I will refactor `root.py` to leverage FastAPI dependency injection as that is the suggested approach in the [documentation](https://fastapi.tiangolo.com/advanced/settings/#creating-the-settings-only-once-with-lru_cache).

backend/api/core/middleware/authorization.py
```diff
@@ -0,0 +1,77 @@
...
+
+            if bool(self.AUTHORIZED_OKTA_GROUPS & set(token_data.groups)):
+                request.state.user = token_data
+                response = await call_next(request)
+            else:
+                raise UnauthorizedOktaGroup()
+        except (UnauthorizedOktaGroup, JWTError) as e:
+            context = (
+                "See README for instructions on creating a JWT token if developing locally"
+                if e.__class__ == JWTError
```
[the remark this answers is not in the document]
drew-beckmen: I had considered this, but I wanted to keep it in a single `except` block given that the HTTP status code is the same, so it avoids repetition.

backend/api/tests/unit/middleware/test_authorization.py
```diff
@@ -0,0 +1,72 @@
...
+    assert (
+        "Ensure you are a member of an authorized Okta group"
+        in response.json()["message"]
+    )
+
+
+def test_valid_jwt():
+    authorizd_client = FastApiTest()
+
+    response = authorizd_client.get("/")
```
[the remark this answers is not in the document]
drew-beckmen: good catch 😅

environments/backend-staging.yaml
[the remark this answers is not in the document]
drew-beckmen: I've added links to the Kanopy portal to the README for future reference.

backend/main.py
```diff
@@ -1,18 +1,17 @@
...
-    allow_methods=["*"],
-    allow_headers=["*"],
-)
 
-app.include_router(hello_world.router)
+@app.on_event("startup")
+async def startup_db_client():
+    client = motor_asyncio.AsyncIOMotorClient(settings.mongo_uri)
+    await init_beanie(
+        database=client[settings.mongo_db_name], document_models=[Entitlement]
```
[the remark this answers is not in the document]
drew-beckmen: Added custom error handler for cases in which `MONGO_URI` or `MONGO_DB_NAME` are invalid. Thanks for the suggestion!

backend/api/tests/unit/middleware/test_authorization.py
```diff
@@ -0,0 +1,72 @@
...
+        "User is not authorized to access this resource." in response.json()["message"]
+    )
+    assert (
+        "Ensure you are a member of an authorized Okta group"
+        in response.json()["message"]
+    )
+
+
+def test_valid_jwt():
+    authorized_client = FastApiTest()
```
rayangler, line 69: Just noticed this now. Why do we declare `authorized_client` both locally in this test and globally? I don't think I see the `authorized_client` used anywhere outside of this method either (I expect we'll plan to use it once we have other authorized routes we'd want to test). Should we stick to one?

rkhullar-mongodb (approved):

backend/api/tests/unit/middleware/test_authorization.py
drew-beckmen: I moved `unauthorized_client`  into the global scope because it was used across multiple tests, and I roped `authorized_client` into that change as well. However, I agree that we should just have the locally scoped variable for the one test in which it is used. I'll remove the global variable at the top of the file. Thanks for catching!

PR #10, opened 2022-07-22
DOP-3152: Set up database for testing

rayangler:
### Ticket

DOP-3152 - Shared prerequisite for endpoint testing.

### Notes

- Just so we have tests to work with in general, I decided to set up basic integration testing with a local database. Testing requires that a local `mongodb` instance is running (see the backend `README` for more details). `pytest` will set up and tear down the testing database accordingly.
- In the future, or alternatively, we can pivot to using unit tests through mocks. We could also potentially look into using transactions and cancelling them to prevent us from needing to actually update any of the sample data generated for `hapley-test`.


.github/workflows/pull-request-checks.yml
```diff
@@ -43,6 +43,8 @@ jobs:
           pipenv install --dev
       - name: Format and lint
         run: pipenv run format && pipenv run lint
+      - name: Start MongoDB
+        run: docker run -d -p 27017:27017 --name test-mongo mongo:latest
```
drew-beckmen, line 47: This approach with Docker works for me, but there is another approach via a [published GH Action](https://github.com/marketplace/actions/mongodb-in-github-actions) if we want to go that route.

backend/api/database.py
```diff
@@ -0,0 +1,24 @@
+from beanie import init_beanie
+from motor import motor_asyncio
+from pymongo.errors import ConfigurationError, InvalidName
+
+from api.core.config import Settings
```
drew-beckmen, line 5: Love this abstracted out into a separate file!

backend/main.py
```diff
@@ -1,27 +1,11 @@
...
-from pymongo.errors import ConfigurationError, InvalidName
-
-from api.core.config import get_settings
+from api.core.config import Settings
 from api.core.factory import create_app
-from api.model.entitlement import Entitlement
+from api.database import start_db_client
 
-settings = get_settings()
+settings = Settings()
```
drew-beckmen, line 5: Is there a reason we're no longer using the `get_settings()` method?

backend/tools/data/entitlements.json
```diff
@@ -0,0 +1,4 @@
+[
+  { "email": "foo@gmail.com", "github_username": "foo_git", "repos": ["docs-landing"], "slack_user_id": "0" },
+  { "email": "test@mongodb.com", "github_username": "test_git", "repos": ["docs", "docs-landing"], "slack_user_id": "1"  }
```
drew-beckmen, line 3: Do you think it makes more sense to include these files in a subdirectory of `tests`?

backend/tools/test_db.py
```diff
@@ -0,0 +1,57 @@
...
+    client = MongoClient(LOCAL_URI)
+
+    if DB_NAME in client.list_database_names():
+        client.drop_database(DB_NAME)
+
+    db = client[DB_NAME]
+    entitlements_coll = db[COLL_ENTITLEMENTS]
+    repos_coll = db[COLL_REPOS_BRANCHES]
+
+    bulk_insert_items(entitlements_coll, "./data/entitlements.json")
```
drew-beckmen, line 41: We could avoid all the code in `bulk_insert_items` by running a `mongoimport` command directly. It's probably not as clean to run a shell command directly but it is more concise.

backend/tools/test_db.py
```diff
@@ -0,0 +1,57 @@
...
+    base_path = Path(__file__).parent
+    file_path = base_path / filename
+    f = open(file_path)
+
+    data = json.load(f)
+    for item in data:
+        writes.append(InsertOne(item))
+
+    result = collection.bulk_write(writes)
+    print(result.bulk_api_result)
```
drew-beckmen, line 24: did you forget to remove this? It clutters the output when running the test suite which isn't ideal.
```
{'writeErrors': [], 'writeConcernErrors': [], 'nInserted': 2, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nRemoved': 0, 'upserted': []}
{'writeErrors': [], 'writeConcernErrors': [], 'nInserted': 2, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nRemoved': 0, 'upserted': []}
```

drew-beckmen (commented):
In addition to the changes you've made thus far, we'll need to update `.drone.yml` to launch a Mongo instance since we run tests on merges into `main` as well.

Overall, I'm pretty satisfied with this approach. Setting up Mongo added 12 seconds to the CI run which isn't too bad. One consideration will be how we want to handle tests that have side effects (`POST`, `PUT`, and `DELETE` routes).  It's possible we'll want to load fresh data prior to every test, but we can assess that at a later point.

.drone.yml
```diff
@@ -11,6 +11,7 @@ steps:
     - cd backend
     - pip install pipenv
     - pipenv install --dev
+    - docker run -d -p 27017:27017 --name test-mongo mongo:latest
```
drew-beckmen, line 14: I think we need to declare MongoDB as a service, which allows it to become available throughout the pipeline: https://blog.ruanbekker.com/blog/2019/05/04/using-mongodb-inside-drone-ci-services-for-unit-testing/

rayangler: Thank you for the heads up! Foolishly thought it would work like in GH Actions. Learned about `services` 😄

backend/tools/test_db.py
drew-beckmen: Sounds good to me! Number 2 is an especially good point

.drone.yml
drew-beckmen: not foolish - I would've done the exact same thing :)

rayangler:
> In addition to the changes you've made thus far, we'll need to update .drone.yml to launch a Mongo instance since we run tests via Drone on merges into main as well.

Thank you for the reminder on this! I tried out the backend pipeline on this branch and it seems like the tests pass successfully.

> One consideration will be how we want to handle tests that have side effects (POST, PUT, and DELETE routes). It's possible we'll want to load fresh data prior to every test, but we can assess that at a later point. In general, we should strive to preserve independence between tests (ie. the side effects of one test should not be able to influence the outcome of a subsequent test)

I agree. I think we'll just need to be mindful of how we write our tests for now until we figure out a more concrete way to ensure that the tests are always independent

.drone.yml
```diff
@@ -4,6 +4,13 @@ kind: pipeline
 type: kubernetes
 name: test-backend
 
+services:
+  - name: test-db
+    image: mongo:latest
+    detach: true
```
drew-beckmen, line 10: based on the [docs](https://docs.drone.io/pipeline/docker/syntax/services/#detached-steps), it seems like `detach` is only relevant when you define the service as a step within your pipeline

.drone.yml
```diff
@@ -4,6 +4,13 @@ kind: pipeline
 type: kubernetes
 name: test-backend
 
+services:
+  - name: test-db
```
drew-beckmen, line 8: Maybe `setup-mongo` is a better name? it shows up as a step in the Drone UI, and I think `test-db` is a bit ambiguous.
Screen Shot 2022-07-25 at 5 09 39 PM

drew-beckmen (approved):
Just a few comments, but no blocking concerns. Thank you for digging into this!

backend/api/tests/base.py
```diff
@@ -1,20 +1,35 @@
...
+            test_username = "foo"
             token = Authorization.build_sample_token(
-                email="foo@gmail.com", username="foo"
+                email="foo@gmail.com", username=test_username
             )
-            self.headers = {"cookie": "auth_user=foo; auth_token=" + token}
+
+            # Set cookies that are automatically set through CorpSecure
+            self.cookies["auth_user"] = test_username
+            self.cookies["auth_token"] = token
```
rayangler, line 32: I had to change how cookies were being set when using `FastApiTest` in the form:

```
with FastApiTest() as client:
    ...
```
For some reason, using the test client like above with the previous iteration was causing the `/repos` `GET` request to go through the authorization middleware twice, where the first pass's cookies were being set by my env vars, and the second pass's cookies were empty.

rayangler (commented):
@drew-beckmen - gonna re-request your review just so you're aware of the recent change to settings cookies on the test client.

I also fixed a small bug I caused by moving the test files around. 😅 This whole time, the integration folder wasn't being picked up by our test command. Should be fine now though!

backend/tools/test_db.py
```diff
@@ -0,0 +1,59 @@
...
+    data = json.load(f)
+    for item in data:
+        writes.append(InsertOne(item))
+
+    collection.bulk_write(writes)
+
+
+# Reminder: Ensure that local database is set up and running first
+# TODO: Set up indexes for each collection.
+def setup_test_db():
```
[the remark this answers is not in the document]
rayangler: Good idea!

backend/api/tests/tools/setup_db.py
```diff
@@ -0,0 +1,62 @@
...
+
+class TestDatabase:
+    def __init__(self):
+        self.client = MongoClient(TEST_LOCAL_DB_URI)
+
+    def _bulk_insert_items(self, collection, filename):
+        writes = []
+
+        base_path = Path(__file__).parent
+        file_path: Path = base_path / filename
```
rkhullar-mongodb, line 20: nice to see pathlib usage!

backend/api/tests/tools/setup_db.py
```diff
@@ -0,0 +1,62 @@
...
+        f = open(file_path.resolve())
+
+        data = json.load(f)
+        for item in data:
+            writes.append(InsertOne(item))
```
rkhullar-mongodb, line 25: nit: For reading the json from the file object, or reading files in general python has a context manager. The connection to the file will automatically close after the block of code within the manager executes, even if an error gets raised:

```python
with file_path.resolve().open('r') as f:
    data = json.load(f)
```

optional nit: For generating the write operations you could use list comprehension:
```python
writes = [InsertOne(item) for item in data]
```

.drone.yml
```diff
@@ -4,6 +4,12 @@ kind: pipeline
 type: kubernetes
 name: test-backend
 
+services:
+  - name: setup-mongo
+    image: mongo:latest
```
rkhullar-mongodb, line 9: nit: should specify a version tag besides `latest`. maybe `mongo:5.0`

rkhullar-mongodb (approved):
Looks good, had a couple optional suggestions

.drone.yml
rayangler: Done. Also updated the version tag on the GH action accordingly

PR #12, opened 2022-07-26
DOP-3132: Add routes for `/repos/{repo_name}/groups`

rayangler:
### Ticket

DOP-3132

### Notes

- There's an open question for how we want to isolate tests as much as possible. This is mostly due to how we only set up and tear down the db when `pytest` starts and when `pytest` ends, respectively (#10). For now, these tests are not completely isolated as we figure out what the best approach is (e.g. set up/tear down the db per test or per test file, explore mocking, etc.)
- A lot of this code may need to get refactored depending on model changes, validations through a separate class vs. dependencies, etc. for the `/versions` endpoints


backend/api/model/repo.py
```diff
@@ -0,0 +1,181 @@
...
+            for group in self.repo.groups:
+                self.validate_one_group(group)
+
+        self.validate_one_group(new_group)
+        self.raise_errors()
+
+
+async def find_one_repo(repo_name: str) -> Repo | None:
+    repo = await Repo.find_one(Repo.name == repo_name)
+    return repo
```
drew-beckmen, line 141: Just an FYI that I plan to move this function to a dependency so all our path operation functions receive a `Repo` rather than the raw string `repo_name`. We can also raise `RepoNotFound` errors here so that we don't have to repeat the code 
```
    if not repo:
        raise RepoNotFound(repo_name)
```

backend/api/tests/integration/routes/test_groups.py
```diff
@@ -0,0 +1,138 @@
...
+
+def get_groups(client: FastApiTest, repo_name: str):
+    response = client.get(groups_route(repo_name))
+    assert response.status_code == 200
+    groups = response.json()["groups"]
+    return groups
+
+
+def test_groups_get():
+    entitled_user = "test@mongodb.com"
```
drew-beckmen, line 16: can you write a test for an `unentitled_user`? I also think we could pull out the `entitled_user` variable into the global scope rather than redefining it in each test

backend/api/tests/integration/routes/test_groups.py
```diff
@@ -0,0 +1,138 @@
...
+        groups = get_groups(client, "docs")
+        assert len(groups) == 2
+
+
+def test_groups_get_none():
+    entitled_user = "test@mongodb.com"
+    with FastApiTest(email=entitled_user) as client:
+        # Nonexistent groups
+        groups = get_groups(client, "docs-landing")
+        assert groups is None
```
drew-beckmen, line 28: I think this is up for debate, but I think it might be more intuitive to expect a response with an empty array rather than `None`. What do you think?

backend/api/tests/integration/routes/test_groups.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def test_groups_post():
+    entitled_user = "test@mongodb.com"
+    with FastApiTest(email=entitled_user) as client:
+        repo_name = "docs"
+        new_group = {"groupLabel": "New Test Group Post", "includedBranches": []}
+
+        groups = get_groups(client, repo_name)
+        assert len(groups) == 2
```
drew-beckmen, line 38: I don't think we need this assertion as we are not testing the `get` route here

backend/api/tests/integration/routes/test_groups.py
```diff
@@ -0,0 +1,138 @@
...
+
+        response = client.post(groups_route(repo_name), json=new_group)
+        res_json = response.json()
+        assert response.status_code == 200
+        assert res_json == new_group
+
+        groups = get_groups(client, repo_name)
+        assert len(groups) == 3
```
drew-beckmen, line 46: In the name of creating tests that avoid hardcoding, could we do:
```
new_groups = get_groups(client, repo_name)
assert len(new_groups) - len(groups) == 1
```

I think the idea of testing "increases the # of group by 1" is a bit cleaner than testing "increases the number of groups to 3"

backend/api/v1/routes/groups.py
```diff
@@ -0,0 +1,33 @@
...
+    dependencies=[Depends(check_if_user_entitled_to_repo)],
+)
+async def read_groups(repo_name: str):
+    return await Repo.find_one(Repo.name == repo_name).project(RepoGroupsView)
+
+
+@router.post(
+    GROUPS_INDEX_PATH,
+    response_model=Group,
+    dependencies=[Depends(check_if_user_entitled_to_repo)],
```
drew-beckmen, line 24: We can move the dependency to the `APIRouter` level since it applies to every path decoration operator. See example [here](https://fastapi.tiangolo.com/tutorial/bigger-applications/#import-the-dependencies).

drew-beckmen (commented):
Looks really solid! Just a couple comments that need addressing, but I left the review with the understanding that my PR that I'll open later today will touch much of this code again.

backend/api/exceptions.py
```diff
@@ -0,0 +1,33 @@
...
+class RepoNotFound(HTTPException):
+    def __init__(self, repo_name: str) -> None:
+        super().__init__(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"The repo {repo_name} does not exist.",
+        )
+
+
+class ValidationError(HTTPException):
+    def __init__(self, message, errors: list[str]) -> None:
```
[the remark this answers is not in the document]
rayangler: Yeah, I think this one is a special case considering that there's a benefit in showing a list of everything that's wrong with groups compared to showing one error at a time. Should we stick to just raising one error at a time instead to keep things uniform?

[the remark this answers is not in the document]
drew-beckmen: I was thinking of having a ErrorResponse model structured as follows:
```
{
   "detail": {
       "message": str
      "errors": list[str]
   }
}
```

This allows us to return multiple errors in the array at one time if necessay. I could see the frontend consuming this data by showing an error banner where the heading is the `message` and then there's a list of `errors` underneath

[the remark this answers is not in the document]
rayangler: Gonna leave this conversation open for now as a general reminder for us/everyone that this is something we *should* do, but don't need right now as we build up respective initial exceptions across multiple PRs. We can address this in a follow-up PR once the model is more concrete

backend/api/model/repo.py
rayangler: Sounds good to me with our current use case. The only reason I left out raising the exception in this function is because not finding a repo, at its core, is not necessarily an error and it allows us to be flexible in terms of what it means for `repo == None`

backend/api/tests/integration/routes/test_groups.py
rayangler: Sure, I can write a test for this. I prefer keeping `entitled_user` in a local scope though. There may be instances where we won't necessarily be working with the same entitled user per test.

backend/api/tests/integration/routes/test_groups.py
rayangler: Keeping it `None` preserves the absence of `groups` from a repo as the data does in the db currently, but I can try adding a validator to the response model for the `GET` request such that `groups` will be an empty list instead of `None`.

drew-beckmen: Ah right forgot about that field being optional! If you can figure out how to return an empty array in the response, I think that would be a lot clearer from a client perspective.

backend/api/tests/integration/routes/test_groups.py
rayangler: Good idea

backend/api/tests/integration/routes/test_groups.py
rayangler: Just pushed up the change for this! Let me know what you think

drew-beckmen: Looks good to me!

drew-beckmen (approved):

PR #13, opened 2022-07-27
DOP-3157: Add unique ids to all subdocuments in repo_branches.branches and repo_branches.groups

drew-beckmen:
### Ticket

[DOP-3157](https://jira.mongodb.org/browse/DOP-3157)

### Overview

The script adds an `id` field of type `ObjectID` to every subdocument in the `branches` and `groups` array fields.

Testing Story: ran against a clone of the `pool.repos_branches` collection

In addition to the script, this PR:

- updates the seed data used for tests
- adds a custom wrapper around `bson.ObjectId` to validate object IDs 
- updates Pydantic models so FastAPI understands how to encode data from our wrapper class as JSON (which is as strings)

- [x] After the script is run, I'll remove it from the PR and just merge in the changes to the test data.


scripts/add_ids.py
```diff
@@ -0,0 +1,68 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-env', '--environment', choices=['dev', 'test', 'prod'], help='Environment to run the script in', default='dev')
+parser.add_argument('--rollback', action='store_true', help="Drop email address field from collection")
+args = parser.parse_args()
+
+COLLECTION_NAME_BY_ENVIRONMENT = {
+    'dev': 'drew_repo_branches',
+    'test': 'repos_branches',
+    'prod': 'entitlements'
```
rayangler, line 19: Wondering why this is `entitlements` and not `repos_branches`

rayangler (commented):
Makes sense to me. Just one question

scripts/add_ids.py
drew-beckmen: because I copy and pasted :( will fix - thanks for catching!

scripts/add_ids.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+from pymongo import MongoClient, collection
+from bson import ObjectId
+from decouple import config
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-env', '--environment', choices=['dev', 'test', 'prod'], help='Environment to run the script in', default='dev')
+parser.add_argument('--rollback', action='store_true', help="Drop email address field from collection")
```
rayangler, line 13: Also just wanted to call out this help message isn't relevant to what the script does

scripts/add_ids.py
```diff
@@ -0,0 +1,68 @@
...
+
+def connect() -> MongoClient:
+    DB_USERNAME = config('DB_USERNAME')
+    DB_PASSWORD = config('DB_PASSWORD')
+    CONNECTION_STRING = f"mongodb+srv://{DB_USERNAME}:{DB_PASSWORD}@cluster0.ylwlz.mongodb.net/snooty_dev"
+    return MongoClient(CONNECTION_STRING, uuidRepresentation='standard')
+
+
+def modify_uuids(collection: collection.Collection, removeIds: bool = False):
+    print("Rollback flag not set. Adding ids to subdocuments in branches and groups fields")
```
rayangler, line 37: One last thing! Idk if this was intended to be under an if statement or something?

scripts/add_ids.py
drew-beckmen: copy and paste strikes again. fixed!

scripts/add_ids.py
drew-beckmen: yep good catch

backend/api/v1/routes/groups.py
```diff
@@ -11,7 +11,7 @@
 
 @router.get(GROUPS_INDEX_PATH, response_model=RepoGroupsView)
 async def read_groups(repo_name: str):
-    return await Repo.find_one(Repo.name == repo_name).project(RepoGroupsView)
+    return await Repo.find_one(Repo.name == repo_name)
```
rayangler, line 14: (General comment for discussion, no change needed) The effects would be minimal given the size of our data, but I'm wondering what the general good practice is for using projections with response models. 

My understanding based on the [Beanie docs](https://roman-right.github.io/beanie/tutorial/finding-documents/#projections) is that projecting can save on the network request and processing for the data returned from the db. Meanwhile, FastAPI's response model will filter the unneeded attributes from the model we're returning to allow it to fit the response model's schema. It feels like we *should* be saving some initial bandwidth through projection, but I'm not 100% certain. Maybe something worth testing in the future

backend/api/model/pydantic_object_id.py
```diff
@@ -0,0 +1,18 @@
...
+class PyObjectId(ObjectId):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        if not ObjectId.is_valid(v):
+            raise ValueError("Invalid object_id")
+        return ObjectId(v)
```
rayangler, line 18: Beanie has a built-in `PydanticObjectId` class that seems to have the same methods and inheritance of `ObjectId`. It also uses some `utf-8` decoding, which might help with translating it to a string type. 

Is that something we can use, or does it not work for our use case?

backend/api/tests/integration/routes/test_groups.py
```diff
@@ -18,6 +18,9 @@ def post_groups(client: FastApiTest, repo_name: str, new_group: dict):
     response = client.post(groups_route(repo_name), json=new_group)
     res_json = response.json()
     assert response.status_code == 200
+    # assigned a valid object id
+    assert len(res_json["id"]) > 0
+    del res_json["id"]
```
rayangler, line 23: Would it make more sense to just add the `id` on `POST`/insert of the new group instead of deleting it for the test? If not, we can handle including the `id` in a follow-up PR.

rayangler (commented):
Looks good right now. Thanks for looking into this!

backend/api/tests/integration/routes/test_groups.py
drew-beckmen: The `id` gets autogenerated upon insertion into the DB and is included in the response to the client, so I'm a bit confused by your question. The test mimics a client submitting a new group, and the client wouldn't be responsible for creating the ID. Basically the test is saying does the server respond with the same data the client sent except for the `id` autogenerated by the server

backend/api/model/pydantic_object_id.py
drew-beckmen: Nice find in Beanie! Let me explore this class!

backend/api/v1/routes/groups.py
drew-beckmen: Thanks for pointing this out in the docs! I was under the (bad) assumption that the projection and setting `response_model` accomplished the same thing, but that's not the case. The projection is changing the amount of data we query at the database level, while `response_model` does the filtering at the application level. Thanks for pointing this out, and I agree that we should leave the projection as a part of the query!

backend/api/tests/integration/routes/test_groups.py
rayangler: Ah, yes, you're right! Sorry about the confusion there

backend/api/model/pydantic_object_id.py
drew-beckmen: This works flawlessly! Removing all the extra code I wrote and leveraging the class from Beanie.

backend/api/v1/routes/groups.py
rayangler: It was a reasonable assumption to make given that the response would ultimately be the same

rayangler (approved):

PR #17, opened 2022-07-29
DOP-3131: `repos/{repo_name}/versions/{version_id}` endpoints

drew-beckmen:
### Ticket

[DOP-3131](https://jira.mongodb.org/browse/DOP-3131)

### Description

This pull request adds three routes at `repos/{repo_name}/versions/{version_id}`, namely, the `GET`, `PUT`, and `DELETE` routes. The `version_id` references the ids added to subdocuments in #13 

Screen Shot 2022-07-29 at 4 41 20 PM

Additionally, this PR fixes a bug in an untested codepath in the `Version` model where the `ValidationError` was raised with a `errors` as a single string rather than a list of strings. It also includes a new parent class called `VersionIn`, which lacks an `id` to ensure that the OpenAPI spec has the correct request bodies. When creating a version, the `id` gets created server side. When modifying an existing version, the `id` gets pulled from the path parameter.


backend/api/models/repo.py
```diff
@@ -59,3 +60,22 @@ async def reorder_repo_version(
     repo.versions.insert(target_index, repo.versions.pop(current_index))
     await repo.update({"$set": {"branches": repo.versions}})
     return repo
+
+
+def get_version_index(repo: Repo, version_id: ObjectId) -> int:
+    return repo.versions.index(
+        next(filter(lambda v: v.id == version_id, repo.versions))
+    )
```
rayangler, line 68: Correct me if I'm misunderstanding these lines of code, but are we going through the versions list to filter out the object with the expected version id, and then using `.index()` on the same list to find that object's position in the list? Would it be faster to use `enumerate`? Something like:

```
for idx, version in enumerate(repo.versions):
    if version.id == version_id:
        return idx
return -1 # For completion
```

backend/api/models/version.py
```diff
@@ -4,8 +4,9 @@
 from api.exceptions import ValidationError
 
 
-class Version(BaseModel):
-    id: PydanticObjectId = Field(default_factory=PydanticObjectId)
+# id should not be listed as a field in request body because it is automatically
+# generated for POST or pulled from the URL path parameter for PUT.
+class VersionIn(BaseModel):
```
rayangler, line 9: For what it's worth, I don't think there's anything wrong with including the id in the body for the docs. We would just need to ensure that including an id that doesn't match the parameter id doesn't overwrite the existing one (this `VersionIn` pattern does accomplish this though!). 

The side effects of having `VersionIn` that might look a little weird at first glance are that there are now 2 `Version`-like models listed in the docs with the only real difference is the existence of the id field, and we now have an inheritance between `VersionIn` and `Version` for the sake of documentation (not sure if there's a functional purpose for it).

The id being listed in both the parameter and body of the object is also not uncommon to have in documentation:
* [MongoDB App Services API](https://www.mongodb.com/docs/atlas/app-services/admin/api/v3/#operation/adminGetCustomResolver)
* [Notion API](https://developers.notion.com/reference/patch-page)
* [YouTube Data API](https://developers.google.com/youtube/v3/docs/playlists/update)

backend/api/dependencies.py
```diff
@@ -46,12 +52,42 @@ async def find_one_repo(repo_name: str) -> Repo:
...
+        return ObjectId(version_id)
+    except errors.InvalidId:
+        raise ValidationError(
+            message="Invalid ObjectId.",
+            errors=[f"{version_id} is not a valid 24-byte hex string."],
+        )
+
+
+async def find_one_version(
+    object_id: ObjectId = Depends(convert_to_object_id),
```
rayangler, line 66: What's the reason for `convert_to_object_id` as a dependency? Does using `PydanticObjectId` as the data type not work for this use case?

backend/api/tests/integration/routes/test_version.py
```diff
@@ -0,0 +1,100 @@
+from api.tests.base import FastApiTest
+
+from .test_versions import generate_random_valid_version
```
rayangler, line 3: (Nit) Would `generate_random_valid_version` be useful to have as a test util function for visibility/general organization? If not, I'm also okay with leaving it where it is!

backend/api/core/metadata/tags.py
```diff
@@ -19,4 +19,8 @@
         "name": "groups",
         "description": "Manage the list of `groups` for a repo",
     },
+    {
+        "name": "version",
+        "description": "Manage a specific version for a repo",
+    },
```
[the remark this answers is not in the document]
drew-beckmen: Nope, I'm good with combining them!

backend/api/dependencies.py
drew-beckmen: You can use `PydanticObjectId`, but the only downside is that it will return an error of the form:
```
{'detail': [{'loc': ['path', 'version_id'], 'msg': 'Id must be of type PydanticObjectId', 'type': 'type_error'}, {'loc': ['path', 'version_id'], 'msg': 'Id must be of type PydanticObjectId', 'type': 'type_error'}]}
```

if the URL parameter is not a valid ObjectID. Totally agree that I should be using `PydanticObjectId` class here, but I think the a nicely formatted error is better than the default validation error.

backend/api/models/repo.py
drew-beckmen: I think the code is actually equivalent because in Python 3, `filter` returns an iterable and doesn't resolve itself until you do something with the iterable. Based on my understanding, the code I have is essentially syntactic sugar for the for loop since the `next` function will return once the first element of the iterable has been resolved. I'll dig deeper to confirm this.

backend/api/v1/routes/version.py
```diff
@@ -0,0 +1,59 @@
...
+    VERSION_INDEX_PATH,
+    response_model=DeleteItemResponse,
+    description="Delete a specific version",
+)
+async def delete_version(
+    version_id: ObjectId = Depends(convert_to_object_id),
+    repo: Repo = Depends(find_one_repo),
+):
+    await delete_repo_version(repo=repo, version_id=version_id)
+    return DeleteItemResponse(success=True, message="Version deleted successfully")
```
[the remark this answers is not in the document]
drew-beckmen: Works for me. I was modeling this on some other APIs I saw documented online (such as Spotify), but I agree that it's not necessary. We can just rely on the status code.

backend/api/models/version.py
drew-beckmen: Ok I'm fine removing the `VersionIn` model for the sake of simplicity. In terms of the functional purpose it serves now, it would strip out any `id` field passed in the payload automatically, so we wouldn't have to manually overwrite the `id` that either:
1. the user passes the `id` field in the payload
2. or the `id` gets automatically generated by the server. (this was what I wanted to avoid with `VersionIn`. Because both `Version` and `Group` have `id: PydanticObjectId = Field(default_factory=PydanticObjectId)` with the `default_factory`, we'll automatically generate an object ID when we specify the payload as `Version` (whereas this is not the case with `VersionIn`)

backend/api/models/repo.py
rayangler: No worries at all! Thanks for the explanation though

backend/api/models/version.py
rayangler: Ah, good point on the functional purpose. Overwriting/reassigning the `id` will happen regardless if we're constructing a full `Version` object off of the keys from `VersionIn` though. On the other hand, preventing a new `id` from being generated through the default factory would be a good pro for keeping the `VersionIn` model 🤔 

I'm of the opinion that we can revisit this at a later date as it seems like an optional optimization that have consequences on our model structures and docs

backend/api/dependencies.py
rayangler: I didn't consider that this validation error would occur like this! I'm okay with keeping this for now if we want to ensure user/developer-friendly error messages in the event that a malformed ID is being passed through.

In the future, I wonder if there's a way we could create our own subclass off of `PydanticObjectId` that can overwrite their validation logic with the one for `convert_to_object_id` when performing an implicit type change. That might save us from needing to declare the dependency. Something to look into at a later point maybe!

backend/api/models/version.py
drew-beckmen: > Overwriting/reassigning the id will happen regardless if we're constructing a full Version object off of the keys from VersionIn though.

Yes, but instead of generating a random ObjectId that we will then overwrite, we can construct the object using the `id` from the URL path parameter.
```
updated_version = Version(**updated_version.dict(), id=version_id)
```

For now, I'll remove it to keep things simple. If it becomes an issue for clients consuming the API, we can reasses.

rayangler (approved):
LGTM! Thanks for working on this

PR #18, opened 2022-08-02
DOP-3133: Implement routes for `/groups/{group_id}`

rayangler:
### Ticket

DOP-3133

### Notes

- Creates routes for `.../groups/{group_id}`


backend/api/core/validators/valid_group.py
```diff
@@ -52,9 +55,7 @@ def validate_group(cls, values):
                 set(new_group.included_branches) & set(flattened_existing_branches)
             )
             if len(previously_used_versions) > 0:
-                err_msg = (
-                    f'Versions {", ".join(previously_used_versions)} {"are" if len(previously_used_versions) > 1 else "is"} already used in another group.'
-                )
+                err_msg = f'Versions {", ".join(previously_used_versions)} {"are" if len(previously_used_versions) > 1 else "is"} already used in another group.'
```
rayangler, line 58: Not sure why this is being forced to reformat via the pre-commit hook

backend/api/v1/routes/group.py
```diff
@@ -0,0 +1,46 @@
...
+async def read_group(group: Group = Depends(find_one_group)):
+    return group
+
+
+@router.put(
+    GROUP_INDEX_PATH, description="Update a specific group", response_model=Group
+)
+async def update_group(
+    updated_group_params: ValidGroup = Depends(updated_group_validator),
+) -> Group:
```
drew-beckmen, line 36: super nit but I noticed this is the only function for which you specified a return type. I don't think it's strictly necessary, especially given the `respond_model` in the decorator, but perhaps it's worth maintaining consistency?

rayangler: Wouldn't consider this super nit since this consistency is definitely good to have! Thanks for catching this!

backend/api/models/repo.py
```diff
@@ -48,6 +50,25 @@ async def reorder_repo_group(repo: Repo, current_index: int, target_index: int)
     return repo
 
 
+async def update_repo_group(repo: Repo, updated_group: Group) -> Group:
+    _, group_index = find_repo_item_in_list(updated_group.id, repo.groups)
+    if group_index == -1:
+        raise GroupNotFound(updated_group.id)
```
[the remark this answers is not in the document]
drew-beckmen: In this case, we'll save a lot of computation by adding the `find_one_group` dependency at the router level because we'll avoid having to instantiate the `ValidGroup` and do the iterations over all the existing groups if a request with the form `PUT /groups/{random_non-existing-group}` comes in.

backend/api/core/validators/valid_group.py
drew-beckmen: I saw this too and wasn't quite sure what was going on 😅

drew-beckmen (commented):
LGTM! Would love to just hear your thoughts on adding `find_one_group` as a dependency at the router level before merging. I agree that someone unaware of the caching mechanisms in place might see the code and interpret it as repetitive computation (despite it being the opposite)

backend/api/utils.py
```diff
@@ -0,0 +1,18 @@
+from typing import TypeVar
+
+from beanie import PydanticObjectId
+
+from api.models.group import Group
```
drew-beckmen, line 5: There is a  pre-existing `util` directory inside `core`. Can we put this code there instead?

backend/api/models/repo.py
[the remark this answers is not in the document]
rayangler: I didn't know about the caching behavior for dependencies. That's really cool, thank you for sharing that!

I agree that using the router-level dependency would be nice to have and would make the 404 reporting a little cleaner. When I tried it yesterday, I only decided against it because I noticed that the `find_repo_item_in_list` util function was being called a couple times, but I guess that's not too big of a deal. I'll make the change later!

backend/api/core/validators/valid_group.py
rayangler: Yeah, I couldn't figure out why. If this happens again, I'll probably have to revisit my personal installation of packages and/or our implementation of hooks

backend/api/utils.py
rayangler: Whoops, totally forgot about this directory. Thanks for flagging. I'll add it there

backend/api/models/repo.py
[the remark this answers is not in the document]
rayangler: I couldn't figure out a clean way to prevent `find_by_obj_id` from being called twice. I was hoping to try something with dependencies but went down a rabbit hole looking into parameterized dependencies. I think we can have this be a known limitation for now and revisit later since it shouldn't necessarily block the implementation of the core logic here. What do you think?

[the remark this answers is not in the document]
rayangler: I did go with using `find_one_group` as a router level dependency

rayangler:
@drew-beckmen Re: the `ObjectId` validation error that we discussed offline, I'll make that into a separate ticket and PR to avoid blocking this one from merging

backend/api/core/util/find_by_obj_id.py
```diff
@@ -8,10 +8,14 @@
 RepoItem = TypeVar("RepoItem", bound=Group | Version)
 
 
+# Known limitation: this util funciton can potentially be called twice per route:
+# Once when checking the id is being used to check for an existing Version/Group,
```
drew-beckmen, line 12: this wording feels a bit awkward. I also think "potentially" is the wrong word as we know the function will be called twice for any successful call to a `PUT` or `DELETE` route.

backend/api/tests/integration/routes/test_group.py
```diff
@@ -0,0 +1,84 @@
+from api.tests.base import FastApiTest
+
+ENTITLED_USER = "test@mongodb.com"
```
drew-beckmen, line 3: This just reminded me that I never added `check_if_user_entitled_to_repo` as a dependency to the version endpoints. I'll create a bug ticket to resolve the issue. Right now, anyone with access to the application could adjust versions for any docs content repo. 😨 

- [ ] @drew-beckmen to create the bug ticket

drew-beckmen (approved):
This looks good! Only major issue I found is a bug that I introduced 😓 

👏 for being basically done with the backend!

backend/api/tests/integration/routes/test_group.py
drew-beckmen: Here's the ticket: https://jira.mongodb.org/browse/DOP-3183

backend/api/core/util/find_by_obj_id.py
rayangler: Good point, I put down "potentially" in the sense that this function is intended to be used for the `GET` route but not won't necessarily be called twice. It'll probably make more sense to make it more explicit though!

Do you have any suggestions to make the wording less awkward? Maybe do something like:

```
Known limitation: this util function is called twice on the `PUT` and `DELETE` routes for individual groups and versions:
1) When checking the if the a Version or Group exists with the given id param
2) When performing the update or deletion method.
```

backend/api/tests/integration/routes/test_group.py
rayangler: Nice catch! I'm thinking I should add a test for that too now, just for completion 😅

backend/api/core/util/find_by_obj_id.py
drew-beckmen: That revised version is great! Maybe we can use the word `executing` instead of `performing` in number 2? Other than that, lgtm

rayangler: Good idea. Done

backend/api/tests/integration/routes/test_group.py
rayangler: Okay, I ended up not including it as a test since you mentioned that it should be included as part of the `groups` route.

PR #19, opened 2022-08-02
DOP-3168: Set up client side routing and basic application shell

drew-beckmen:
### Ticket

[DOP-3168](https://jira.mongodb.org/browse/DOP-3168)

### Description

- Adds a basic application shell with the LG Side Nav component
- Leverages `reach/router` for dynamic client-side routing for individual docs repos
- Side nav is populated with the current user's entitled repositories.
- Adds dev dependencies needed to resolve `reach/router` in test files (Gatsby comes with a forked version of `reach/router`. The vanilla `reach/router` isn't yet compatible with React 18, which raised other errors resolved by adding the `parcel` package. Installs must also happen with `--legacy-peer-deps`.

Screen Shot 2022-08-02 at 6 12 15 PM


frontend/package.json
```diff
@@ -34,17 +34,30 @@
...
     "gatsby-plugin-emotion": "^7.17.0",
     "react": "^18.1.0",
-    "react-dom": "^18.1.0"
+    "react-dom": "^18.1.0",
+    "react-helmet": "^6.1.0",
+    "stream-browserify": "^3.0.0"
   },
   "devDependencies": {
     "@emotion/eslint-plugin": "^11.7.0",
+    "@parcel/namer-default": "^2.6.2",
```
drew-beckmen, line 54: this dependency makes me sad 😢 but I couldn't figure out a way to resolve the errors I was getting when only installing `@reach/router` in `devDependencies` without it

frontend/src/components/SideNav.tsx
```diff
@@ -0,0 +1,83 @@
...
+const sideNavStyling = () => LeafyCSS`
+  height: 100%;
+`;
+
+const titleStyle = LeafyCSS`
+  color: ${palette.black};
+  font-style: normal;
+  font-weight: 600;
+  font-size: 16px;
+  line-height: 28px;
```
drew-beckmen, line 24: I feel like there has to be a better way than copying everything from Figma, but I'll come back to styling in a follow-up.

frontend/package.json
mmeigs: Looks like this bug was possibly fixed in Gatsby 4.18.2 - (https://github.com/gatsbyjs/gatsby/issues/35952) 
Don't know if this is an update we want to do, but worked for me when I was having this issue.

PR #20, opened 2022-08-03
DOP-3168: Set up client side routing and basic application shell

drew-beckmen:
### Ticket

[DOP-3168](https://jira.mongodb.org/browse/DOP-3168)

### Description

- Adds a basic application shell with the LG Side Nav component
- adds `gatsby-browser` and `gatsby-ssr` to render the application wrapped in a `Layout` component. The application will conditionally render the side nav based on the current URL. As long as we navigate internally within the application, the navigation component will not remount, and it should therefore not send additional requests to the backend to fetch a user's entitled repos.
- Side nav is populated with the current user's entitled repositories.
- Client side routing is achieved with the Gatsby [File System API](https://www.gatsbyjs.com/docs/reference/routing/file-system-route-api/#syntax-client-only-routes)
- Update `tsconfig` so Jest works correctly with Typescript
- Updates Jest config to account for mapping between Gatsby's custom reach/router and generic @reach/router import
- Mocks Gatsby Link component, which would otherwise cause errors when testing components that use links like SideNav

### Screenshot 

Screen Shot 2022-08-04 at 11 19 11 AM


frontend/src/pages/repos/[repoName].tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface RepoProps {
+  params: {
+    repoName: string;
+  };
+}
+
+function RepoPage({ params }: RepoProps) {
+  return (
+    <AppShell>
```
Comment (drew-beckmen) [inline frontend/src/pages/repos/[repoName].tsx:13]: I don't love how this is rendered in both `index.tsx` and `[repoName].tsx`, but based on the designs, we'll have pages such as `pages/repos/[repoName]/versions/[versionId]` where we don't want to display the side nav. Does anyone know of a better pattern for this?

frontend/tsconfig.json
```diff
@@ -33,8 +33,13 @@
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
-    // "typeRoots": [],                                  /* Specify multiple folders that act like `./node_modules/@types`. */
-    // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
+    "typeRoots": [
+      "node_modules/@types"
+    ] /* Specify multiple folders that act like `./node_modules/@types`. */,
+    "types": [
+      "jest",
```
drew-beckmen, line 40: needed for Jest to work with TS in VSCode IDE

rayangler (commented):
Just a quick look rn. Will take a closer look later

frontend/gatsby-node.js
```diff
@@ -0,0 +1,20 @@
+const webpack = require('webpack');
+
+// Resolves a know issue with incompatibility between Emotion and Webpack 5
```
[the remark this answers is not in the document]
drew-beckmen: Based on [this slack thread](https://mongodb.slack.com/archives/CM90JQHEJ/p1645813751369449?thread_ts=1645727542.936729&cid=CM90JQHEJ), it seems the underlying issue is with Emotion.

frontend/src/pages/repos/[repoName].tsx
Comment (mmeigs) [reply frontend/src/pages/repos/[repoName].tsx:13]: Co-sign with Raymund here. Interested in watching `wrapPageElement` work.
Thought experiment for the hell of it - probably not worth the trade-offs if I'm understanding correctly - but could there be an argument for `versions` to be its own page/route?

Comment (drew-beckmen) [reply frontend/src/pages/repos/[repoName].tsx:13]: Great call @rayangler and @mmeigs! I have used the `wrapPageElement` function in my original PR at first, but later deleted it. I followed a similar pattern to the SO post above, and it seems to be working pretty well. If the user navigates between pages that all display the side nav, we don't send additional requests to the backend (I think this is actually due to Gatsby's `Link` component ). However, if the user goes to a page directly or refreshes the browser, React triggers a re-render which sends a request to the backend.

mmeigs (approved):
LGTM :)

frontend/src/components/Layout.tsx
```diff
@@ -0,0 +1,39 @@
...
+import ContentContainer from './ContentContainer';
+import SideNav from './SideNav';
+import SiteMetadata from './SiteMetadata';
+import '../styles/global.css';
+
+const GlobalGrid = styled('div')`
+  display: grid;
+  grid-template-areas: 'sidenav contents';
+  grid-template-columns: auto 1fr;
+  grid-template-rows: auto 1fr;
```
rayangler, line 12: (Super nit) I think we can remove rows for now since I don't think we actually have rows in the grid with this layout

frontend/src/components/ContentContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+            font-family: ${fontFamilies.default};
+            font-weight: 400;
+            font-size: 34px;
+            line-height: 39.1px;
+          }
+        `}
+      />
+      <main
+        css={css`
+          margin: 72px 72px 0 72px;
```
rayangler, line 20: This might be a big assumption, but it might make sense to just make all of the margins `72px`. Otherwise, we risk having content touching the edge of the bottom

rayangler (commented):
Looks good overall. Just wanted to flag a couple of things

frontend/package.json
```diff
@@ -34,19 +34,29 @@
...
+    "stream-browserify": "^3.0.0"
   },
   "devDependencies": {
     "@emotion/eslint-plugin": "^11.7.0",
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^13.3.0",
-    "@types/jest": "^28.1.4",
+    "@types/jest": "^28.1.6",
     "@types/node": "^17.0.45",
+    "@types/reach__router": "^1.3.10",
```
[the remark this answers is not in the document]
drew-beckmen: Good call. I don't think it's necessary anymore, but if someone ever does leverage elements of the Reach Router API that ships with Gatsby, we'll need to add the package back in to satisfy TypeScript. I'll remove it for now.

frontend/src/components/ContentContainer.tsx
drew-beckmen: That sounds fair to me!

rayangler (commented):
Not sure why this is happening but there's a margin on the side nav title that's causing the top border line below it from extending the full width of the side nav. Is this showing up locally for you as well? Not sure if it's just a rendering issue on my end or something

Margin:
Screen Shot 2022-08-04 at 1 36 33 PM

Border:

Screen Shot 2022-08-04 at 1 44 30 PM

mmeigs:
> 

Looks like it's the max-width property that's stopping it.

frontend/src/components/Layout.tsx
```diff
@@ -0,0 +1,39 @@
...
+`;
```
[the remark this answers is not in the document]
drew-beckmen: I just pushed a change that resolves this issue. Thanks for pointing out the LG recommendation — that was super helpful! Now, if the page content overflows the viewport but the nav bar does not, only the page content will be scrollable. If the nav bar overflows 100vh, it also becomes scrollable. This prevents the "collapsible" icon from being pushed outside the viewport.

drew-beckmen:
> > 
> 
> Looks like it's the max-width property that's stopping it.

Nice catch @rayangler and yep @mmeigs is 100% correct. We actually don't need that max-width CSS property at all and should instead be setting the padding such that the width resolves itself given the width of the parent container (the side nav). I've removed the `max-width` and adjusted the padding such that the spacing is maintained and the border extends to the end of the side nav. Thanks all for catching!

rayangler (commented):
Just wanted to flag one edge case and one question about a specific pattern/decision!

Also, would it be a good idea to have a unit test for the `SideNav` component with some mocked data in place of the fetch? Since we have some logic to determine what's being rendered there, it might be nice to have a basic test to ensure everything is being rendered the way we want it to. This can be handled separately, if we want to merge this PR in ASAP to unblock anything else.

frontend/src/pages/repos/[repoName].tsx
```diff
@@ -0,0 +1,14 @@
...
+import Repo from '../../components/Repo';
+
+interface RepoProps {
+  params: {
+    repoName: string;
+  };
+}
+
+function RepoPage({ params }: RepoProps) {
+  return <Repo repoName={params.repoName} />;
```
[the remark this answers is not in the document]
Comment (drew-beckmen) [reply frontend/src/pages/repos/[repoName].tsx:11]: My primary reason for mapping the pages to components directly was because it allows us to avoid the messiness of routing logic in the test environment. When I made my original attempt at client side routing by serving the app out of a single page, testing became nightmarish because the `reach/router` that ships with Gatsby is a modified fork of the `reach/router` package that is generally available. Since Gatsby is not available in the test environment, I had to install the regular `reach/router` in devDependencies, which introduced a host of other issues with inconsistencies in version requirements.

frontend/src/components/SideNav.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </SideNavItem>
+      <SideNavItem className={cx(navItemStyle(location.pathname === '/'))} as={Link} to="/">
+        Getting Started
+      </SideNavItem>
+      {entitledRepos.length > 0 &&
+        entitledRepos.map((repo) => {
+          const repoName: string = repo.split('/')[1];
+          return (
+            <SideNavItem
+              className={cx(navItemStyle(location.pathname.includes(repoName)))}
```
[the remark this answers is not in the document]
drew-beckmen: Nice catch! I've updated the logic to better handle these edge cases:
```
  const isActive = (repoName: string): boolean => {
    const currentDocsProperty = location.pathname.split('/')[2];
    return currentDocsProperty === repoName;
  }
```

frontend/src/pages/repos/[repoName].tsx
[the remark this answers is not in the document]
Comment (rayangler) [reply frontend/src/pages/repos/[repoName].tsx:11]: Gotcha, using it for testing makes sense. Thanks!

frontend/tsconfig.json
drew-beckmen: I'm actually still having issues with Intellisence complaining in VSCode... Anyone else having this issue?

drew-beckmen: To be more specific, I've isolated the issue to the monorepo setup. If I open just the `frontend` folder in VSCode, the IDE understands and provides completion for Jest tests written in the `tests` directory. However, if I open the entire `docs-hapley` directory in VSCode, VSCode can't properly source Jest. For example:
Screen Shot 2022-08-08 at 8 01 52 PM

frontend/tests/unit/utils/mock-hapley-api.ts
```diff
@@ -0,0 +1,12 @@
+import generateJsonPromise from './generate-json-promise';
+import { GET_REPOS } from './mockData/repos';
+
+export const mockHapleyApi = (url: string) => {
+  const pathname = url.split('/')[1];
+  switch (pathname) {
+    case 'repos':
```
rayangler, line 7: I like the pattern that we're establishing here to help ensure that we're using the correct/expected endpoint for a fetch. It'll be interesting to see how we build this out with other endpoints

frontend/tsconfig.json
rayangler: Nice find regarding the monorepo setup. I've had weird Typescript errors appear/disappear locally as well (although not from the test files)

rayangler (approved):
LGTM! Thanks for taking the time to get this foundation set up

frontend/tsconfig.json
drew-beckmen: I've found that restarting the TypeScript server in VS Code can often do the trick.

PR #21, opened 2022-08-05
DOP-3183: Prevent users from modifying versions on repos to which they are not entitled

drew-beckmen:
### Ticket

[DOP-3183](https://jira.mongodb.org/browse/DOP-3183)

### Notes

- adjusts the base testing client to use an authorized user to prevent repetitive passing of `email` argument to the testing client
- ensures that users can only access `/versions` routes on repos to which they are entitled
- adds `find_one_repo` to the dependencies of the `groups` router. Right now, if a user searches for groups for a repo called `docss`, the error response would indicate that they do not have access (401 unauthorized). However, this is suboptimal and a more helpful error response is a not found (404) error. 
- also fixes an issue where `check_if_user_entitled_to_repo` throws an error if a user has 0 entitlements


mmeigs (approved):
Looks great - especially the test email changes.

rayangler (approved):

PR #24, opened 2022-08-09
DOP-3171: Repo Show Page

drew-beckmen:
### Ticket

[DOP-3171](https://jira.mongodb.org/browse/DOP-3171)

### Description

Adds `/repo/{repo_name}` pages to the Hapley frontend. Displays tables with a docs repo's versions and groups per the design.

### Screenshots
Screen Shot 2022-08-10 at 9 45 04 AM

Screen Shot 2022-08-10 at 9 45 18 AM


frontend/src/components/VersionRow.tsx
```diff
@@ -0,0 +1,51 @@
...
+const VersionActiveContainer = styled('div')`
+  display: flex;
+  padding-left: 24px;
+  align-items: center;
+  justify-content: center;
+`;
+
+const VersionRow = (props: VersionRowProps) => {
+  const { data } = props;
+  const isEmptyRow = data.gitBranchName === '';
```
mmeigs, line 24: Could this be something that's already handled on the backend? Maybe I'm not understanding what edge case this is handling - but I can't think of an instance for this.

mmeigs (commented):
Looks good. We should just keep note of the edge case we found to fix on the backend dealing with branches being an array with a null element. 

- Also, two very odd edge cases to possibly handle that I found because of errors in my entitlements: 
1. someone having a duplicate entitlement. I have two "docs-compass" entitlements and this doesn't break anything, but it's not a pleasing UI. Maybe we can filter these out at some point. 
2. someone having an entitlement that is spelled wrong or invalid. This breaks the UI. At this point, without a github API in our hands, maybe we can add a catch for these failures.

frontend/src/components/VersionRow.tsx
drew-beckmen: It's not handling anything because it's unused code :) I've removed it - nice catch!

frontend/src/components/Repo.tsx
```diff
@@ -1,9 +1,29 @@
...
+import { css, cx } from '@leafygreen-ui/emotion';
+import RepoItems from './RepoItems';
 
-const Repo: React.FC<{ repoName: string }> = ({ repoName }) => (
-  <div>
-    <h1>Accessing data for {repoName}</h1>
-  </div>
-);
+const Repo: React.FC<{ repoName: string }> = ({ repoName }) => {
+  const prettyRepoName = repoName
```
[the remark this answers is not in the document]
drew-beckmen: I agree that we may ultimately want to move this to the database level, but I'll pull the logic into a shared util function so that we can reuse it across components!

drew-beckmen:
> * Also, two very odd edge cases to possibly handle that I found because of errors in my entitlements:
> 
> 1. someone having a duplicate entitlement. I have two "docs-compass" entitlements and this doesn't break anything, but it's not a pleasing UI. Maybe we can filter these out at some point.

I agree that this is not ideal, but I don't anticipate it being an issue. I also expect these entitlements to either become managed through Hapley or via Mana at some point in the not too distant future. @rayangler what do you think?

> 2. someone having an entitlement that is spelled wrong or invalid. This breaks the UI. At this point, without a github API in our hands, maybe we can add a catch for these failures.

This is something we should handle in the form of more generic `catch` blocks on the frontend to catch any unexpected server errors. I'll adjust the `fetch` calls to include a `catch` block!

rayangler:
> I agree that this is not ideal, but I don't anticipate it being an issue. I also expect these entitlements to either become managed through Hapley or via Mana at some point in the not too distant future. @rayangler what do you think?

I think it's fine to add some logic to ensure that we're only displaying an entitled docs repo only once here. I haven't encountered this locally with our backend + frontend yet, but there is one edge case where we could have multiple of the same docs repos, but different forks in a list of entitlements. This use case was mostly for test-deploying forks of docs repo for obscure/specific/experimental changes.

drew-beckmen:
> Note that I'm okay with handling this in a separate PR since this was something we should've caught with the introduction of the side nav component, and isn't necessarily in the scope of the changes of this ticket.

Created a small 1 point ticket in the backlog to handle this filtering out of duplicate repositories on the frontend: https://jira.mongodb.org/browse/DOP-3198

frontend/tests/unit/RepoItemHeading.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+describe('RepoItemsHeading', () => {
+  it('renders heading with pretty format', () => {
+    const wrapper = render(<RepoItemsHeading itemType="version" />);
+    expect(wrapper.getByText('Versions'));
+  });
+
+  it('renders a link to create new version', () => {
+    const { getByTestId } = render(<RepoItemsHeading itemType="version" />);
+    const newVersionLink = getByTestId('newRepoItem');
+    expect(newVersionLink.children[0].textContent).toBe('New Version');
```
rayangler, line 14: Is having a `data-testid` on the `Link` component necessary here? RTL recommends that test id should only be used as a last resort if the other queries don't work for our use case.

Would this test still work if we were to get the button by its content "New Version"?

frontend/src/components/RepoItemsHeading.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  // TODO: update the link once pages are created
+  return (
+    <HeadingContainer>
+      <h4>{`${prettyItemName}s`}</h4>
+      <Link to="/" style={{ textDecoration: 'none' }} data-testid="newRepoItem">
+        <Button variant="primary" size="large" leftGlyph={<Icon glyph="Plus" />}>
+          New {prettyItemName}
+        </Button>
+      </Link>
```
rayangler, line 24: Could this be simplified using the `as` prop? This is something we do on snooty but wondering if this could be applied here as well:

```
<Button as={Link} /* other props */ >
  New {prettyItemName}
</Button>
```

rayangler (commented):
Great work on this! Just flagging suggestions and nits to help make things cleaner/consistent. 

Something in particular that caught my eye was that the `RepoItems` table component can potentially be more generalized to make it reusable for not only the repo show page, but for also the group form for when draggable rows are shown for included versions. If this doesn't seem particularly useful or good of an idea, feel free to ignore since (1) it might make more sense to use a completely separate way to handle the table for versions in the group form, and (2) it might be a big/annoying change to make in a small amount of time for no immediate benefit.

frontend/src/components/GroupRow.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { TableCell } from './RepoItems';
+import { GroupProperties } from '../interfaces/group';
+import { Cell } from '@leafygreen-ui/table';
+import RepoItemRow from './RepoItemRow';
```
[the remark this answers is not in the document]
drew-beckmen: I previously made [this backlog ticket](https://jira.mongodb.org/browse/DOP-3191) to add an ESLint plugin for import ordering, but I've manually reordered the imports in this PR to ensure internal imports are at the bottom.

frontend/src/components/RepoItems.tsx
```diff
@@ -0,0 +1,102 @@
...
+import VersionRow from './VersionRow';
+import GroupRow from './GroupRow';
+import { VersionProperties } from '../interfaces/version';
+import { GroupProperties } from '../interfaces/group';
+
+export type RepoItemProperties = VersionProperties | GroupProperties;
+
+export const TableCell = styled(Body)`
+  padding-left: 24px;
+`;
```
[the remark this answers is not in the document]
drew-beckmen: Sounds good!

frontend/src/components/RepoItems.tsx
```diff
@@ -0,0 +1,102 @@
...
+    fetch(`${process.env.GATSBY_API_URL}/repos/${repoName}/${itemType}s`)
+      .then((res) => res.json())
+      .then((data) => {
+        setItems(data[jsonKey]);
+        setIsLoading(false);
+      });
+  }, [repoName]);
+
+  const getTableHeadings = () => {
+    const tableHeadings = isVersion ? VERSION_COLUMN_NAMES : GROUP_COLUMN_NAMES;
```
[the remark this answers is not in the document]
drew-beckmen: makes sense to me!

frontend/src/interfaces/group.ts
```diff
@@ -0,0 +1,5 @@
+export interface GroupProperties {
```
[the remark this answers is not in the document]
drew-beckmen: yeah that makes a lot more sense :)

frontend/src/interfaces/group.ts
```diff
@@ -0,0 +1,5 @@
...
+  id: string;
```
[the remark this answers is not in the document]
drew-beckmen: I think optional makes sense! Any modifications/additions to the base `Group` and `Version` interfaces can be made through `extend`

frontend/src/interfaces/version.ts
```diff
@@ -0,0 +1,10 @@
+export interface VersionProperties {
+  id: string;
+  gitBranchName: string;
+  active: boolean;
+  urlAlias: Array<string>;
+  publishOriginalBranchName: boolean;
+  urlSlug: string;
+  versionSelectorLabel: string;
+  isStableBranch: boolean;
```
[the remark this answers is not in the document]
drew-beckmen: Could you clarify what you mean here?

frontend/src/components/RepoItemsHeading.tsx
drew-beckmen: We can't pass the Gatsby `Link` component directly as it will throw a TypeScript error. In Snooty, we also have a custom `Link` component that serves as a wrapper.

frontend/src/interfaces/version.ts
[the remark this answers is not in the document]
rayangler: Ah, I mean, given how some of these fields are optional in our [backend models](https://github.com/mongodb/docs-hapley/blob/04ea3a4d3d143f84198d8405febfad38262a5a11/backend/api/models/version.py#L8-L15), should we follow suit to make them optional in our frontend schemas as well? I'm wondering if there's a risk of having an object with an unpopulated field in our backend passed to our frontend, and then having an unexpected type error

frontend/src/components/RepoItemsHeading.tsx
rayangler: Ah, TypeScript 😔 Passing in Gatsby's `Link` component (regardless of it being wrapped) should work fine, but I forgot that LG expects an intrinsic element. Thanks for looking into it anyways!

frontend/src/components/RepoItems.tsx
```diff
@@ -0,0 +1,102 @@
...
+              key={itemType}
+              className={cx(css`
+                height: 56px;
+              `)}
+            >
+              {getTableHeadings()}
+            </HeaderRow>
+          ]}
+        >
+          {({ datum }) => getTableRow(datum)}
```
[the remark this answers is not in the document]
drew-beckmen: I'm done a full refactor to make `RepoItems` a reusable component. I've renamed it `Table`, and now all of the version/group specific logic is pulled into `Repo`.

frontend/src/interfaces/version.ts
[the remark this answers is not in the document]
drew-beckmen: Yeah that makes sense to me!

frontend/src/components/Repo.tsx
```diff
@@ -1,9 +1,93 @@
...
+          (versions.length > 0 ? (
+            <Table<Version> columns={VERSION_COLUMN_NAMES} data={versions} row={getVersionRow} />
+          ) : (
+            emptyState('version')
+          ))}
+      </div>
+      <div>
+        <RepoItemHeading itemType="group" />
+        {!isLoading &&
+          (versions.length > 0 ? (
```
mmeigs, line 83: Should this be groups.length > 0?

frontend/src/components/Repo.tsx
```diff
@@ -1,9 +1,93 @@
...
+];
+export const GROUP_COLUMN_NAMES = ['Name', 'Number of Versions'];
+
+const fetchRepoItemData = (repoName: string, itemType: string, jsonKey: string) => {
+  return fetch(`${process.env.GATSBY_API_URL}/repos/${repoName}/${itemType}`)
+    .then((res) => res.json())
+    .then((data) => data[jsonKey]);
+};
+
+const emptyState = (itemType: string) => {
```
mmeigs, line 29: Such a nit - just to align the semantics, maybe we can name this getEmptyState?

mmeigs (commented):
Such a great change - kudos to both you and @rayangler . Part of me thinks that VersionRow and GroupRow could possibly be combined as well, but I don't see a huge gain from that - plus it's your last day!!! :( Also, you've created such a clean basis for us. It's fantastic.

frontend/src/components/Repo.tsx
drew-beckmen: oh man - that's what you call a copy and paste bug 😱... i've fixed this, but the fact that it slipped through is a good indication that the testing coverage isn't good enough. I'll work on adding some more tests, specifically for these empty states.

frontend/src/components/Repo.tsx
drew-beckmen: sounds good to me!

frontend/src/components/Repo.tsx
drew-beckmen: Just pushed some more tests to improve coverage!

frontend/src/components/Repo.tsx
```diff
@@ -1,9 +1,93 @@
...
+        className={cx(css`
+          margin-bottom: 32px;
+        `)}
+      >
+        {repoDisplayName(repoName)}
+      </H3>
+      <div>
+        <RepoItemHeading itemType="version" />
+        {!isLoading &&
+          (versions.length > 0 ? (
```
rayangler, line 74: Encountering an error here where `versions` is undefined when going on a nonexistent docs repo (such as `/repos/foo`). 

Wondering if we should tweak this conditional, or just redirect to a 404 in this case

drew-beckmen: We could do `versions?.length` and `groups?.length` to avoid the error; however, that then renders the page like this:
Screen Shot 2022-08-16 at 10 30 38 AM
 
I don't like that all appears "normal" here, meaning users could potentially click "New Version" or "New Group". In my opinion, redirecting to a 404 page makes the most sense. I'll add that logic to the fetch logic within `Repo.tsx` for now knowing that it will ultimately be refactored into the hook we create for getting data.

frontend/src/components/Repo.tsx
```diff
@@ -1,9 +1,93 @@
 import React from 'react';
+import { useState, useEffect } from 'react';
+import { H3 } from '@leafygreen-ui/typography';
+import { Body } from '@leafygreen-ui/typography';
+import { palette } from '@leafygreen-ui/palette';
+import { css, cx } from '@leafygreen-ui/emotion';
+import Table from './Table/Table';
```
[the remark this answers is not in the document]
drew-beckmen: sounds good!

frontend/src/components/Repo.tsx
drew-beckmen: @rayangler what sort of pattern do you want to establish here? I think this could also be the domain of a separate ticket responsible for refactoring all data fetching into a shared hook. Are you ok with the `?` conditional navigation operator for now?

rayangler: Sure, I'm okay with `?` for now if we want to spend some time thinking more about the ideal way of handling this separately.

I'm also wondering if this would be a good use case for fetching the data through SSR instead 🤔 But this is more of an open question to better understand what the ideal pattern here would be

drew-beckmen: I decided to add a basic 404 page because we really shouldn't allow users to think a repo `foo` to have versions or groups if the repo doesn't exist. But agreed that this is a larger discussion that needs to happen (likely after I'm gone)

Screen Shot 2022-08-16 at 11 04 29 AM

rayangler (approved):
LGTM! Thanks for all of the hard work. Feel free to merge when @mmeigs has approved as well

mmeigs (approved):

PR #26, opened 2022-08-12
DOP-3173: Implement basic functionality of the Group Form

rayangler:
### Ticket

DOP-3173

### Notes

- There are a lot of pieces that went into making this Group Form component. I tried to make it modular enough with reusable components and hooks. Certain components/hooks (notably, the `useHapleyApi` hook) may need to be consolidated after merging a bunch of our currently open PRs. I'm open to resolving conflicts on this PR to prevent any blocking.
- This PR shares components and logic with #30. We'll work together on consolidating them.


mmeigs (commented):
Got halfway through, looks really good. We divided these tickets relatively poorly, though. We've done very similar work. It will be nice to integrate our changes. Will finish review later!!

rayangler:
> Got halfway through, looks really good. We divided these tickets relatively poorly, though. We've done very similar work. It will be nice to integrate our changes. Will finish review later!!

No worries, this is still in draft in the meantime anyways! I can go ahead and re-request your review when it's officially time.

And yeah, I briefly mentioned if we wanted to have a separate ticket for creating shared form components when we were initially dividing up the frontend work to avoid this issue 😅 . It is easier to see shared patterns between specific components when actually working with them though. We can definitely discuss offline or in a meeting which shared components we'd like to keep so that we're on the same page merging these!

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -0,0 +1,210 @@
...
+  });
+
+  return usedVersions;
+};
+
+const initializeAvailableVersionOptions = (groups: Group[], versions: Version[]) => {
+  const usedVersions = getUsedVersionsInGroups(groups);
+  const unusedOptions: string[] = [];
+
+  versions.forEach((version) => {
```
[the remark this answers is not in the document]
rayangler: Sure, that makes sense to me. I will make the change

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -0,0 +1,210 @@
...
+      unusedOptions.push(versionName);
+    }
+  });
+
+  return unusedOptions;
+};
+
+const GroupForm: React.FC<GroupFormProps> = ({ heading, repoName, group }) => {
+  const groupsEndpoint = `/repos/${repoName}/groups/`;
+  const groups: Group[] = useHapleyApi(groupsEndpoint)?.groups;
```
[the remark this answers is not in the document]
rayangler: Saving on fetches sounds good to me, but I'm not fully aware of how it works with that pattern atm! Mind if we go over this pattern together offline next week with an example?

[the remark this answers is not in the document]
mmeigs: Sure! [Here's the docs](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#pass-state-as-props-to-the-linked-page), and it works like a charm for me so far!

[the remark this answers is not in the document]
rayangler: Update: this was discussed offline. Passing in the data through props/location works great for the generally expected user flow: repos page -> click to go to form -> user is at the form with data populated. This does not work if a user decides to go directly to a form page without needing to perform redirects, as the data will be undefined. We'll revisit optimizing data fetching at a later time

frontend/__mocks__/gatsby.js
```diff
@@ -3,22 +3,37 @@ const gatsby = jest.requireActual('gatsby');
...
+          onClick(e);
+        }
+      };
+
+      return React.createElement('a', {
+        href: to,
+        onClick: onClickOverride,
+        ...rest
+      });
+    }
```
rayangler, line 37: Adding this override logic to prevent the mentioned error in the console. This seems to be happening after changing the `Link` in `RepoItemHeading` to go to the new form link. The relevant line causing this issue is the [link click in the RepoItemHeading test](https://github.com/mongodb/docs-hapley/blob/e06cd63cc00c41df00a6c929856df3a85bb2c660/frontend/tests/unit/RepoItemHeading.test.tsx#L14).

There's an open question as to whether we should even have this override just for the sake of that one test (one test, for now, but can potentially be more in the future?), or if we should modify the test to check for that one behavior differently.

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -0,0 +1,205 @@
...
+      </FormSection>
+      {isEditMode && (
+        <FormFooter>
+          <Button variant="primary" size="large" onClick={addDocument}>
+            Update Document
+          </Button>
+          <Button size="large" onClick={cancelChanges}>
+            Cancel
+          </Button>
+        </FormFooter>
```
rayangler, line 199: This could potentially be its own component since the version form has somewhat similar logic (we could just pass in onClick functionality as props). I wasn't sure at the time of originally writing this if this was preferred though. If so, I can make a new component!

frontend/tests/utils/mock-hapley-api.ts
```diff
@@ -1,22 +1,27 @@
...
-      return generateJsonPromise(getVersions('docs'));
-    case 'repos/docs/groups':
-      return generateJsonPromise(getGroups('docs'));
-    case 'repos/docs-compass/versions':
-      return generateJsonPromise(getVersions('docs-compass'));
-    case 'repos/docs-compass/groups':
-      return generateJsonPromise(getGroups('docs-compass'));
-    default:
-      return Promise.reject(new Error(`Unsupported url: ${url}`));
+export const mockHapleyApi = (url: string): any => {
```
rayangler, line 5: Updated the logic here to make it to more accurately mock the hook itself, rather than the `fetch` within the hook. 

I think there's room for improvement to make this change cleaner, but I also don't want to spend so much time on making this perfect at the expense of holding up the rest of the changes.

rayangler (commented):
Side commentary/annotations

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const cancelChanges = useCallback(() => {
+    navigate(backButtonLink);
+  }, []);
+
+  return (
+    <FormWrapper>
+      <FormHeader>
+        {groupExists && <BackButton text={backButtonText} to={backButtonLink} />}
+        <h1>{heading}</h1>
+        <p>
```
mmeigs, line 139: nit/question - I am new to the group and to Leafygreen. Should we always lean towards using Body or is there a meaningful difference between Body and p? Same goes for h1 - although I'm definitely guilty of this now that I think of it. Mainly a question for me because I currently have a component that should also work for you for this header!

frontend/src/components/Forms/styles.tsx
```diff
@@ -0,0 +1,43 @@
+import styled from '@emotion/styled';
+import { css as LeafyCSS } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+
+const FORM_SECTION_BORDER = `1px solid ${palette.gray.light2}`;
+const FORM_PADDING_LEFT_RIGHT = '72px';
+
```
mmeigs, line 7: Thought for future: should we add a ticket in backlog to make the style more responsive? It does well down about until 900px-1100px depending on the page.

frontend/src/pages/repos/[repoName]/groups/[groupId].tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Group } from '../../../../interfaces/group';
+
+type ServerDataProps = {
+  group: Group;
+  repoName: string;
+};
+
+const GroupPage = ({ serverData }: PageProps<undefined, undefined, undefined, ServerDataProps>) => {
+  const { group, repoName } = serverData;
+  return <GroupForm heading={`Group ${group.groupLabel}`} repoName={repoName} group={group} />;
```
Comment (mmeigs) [inline frontend/src/pages/repos/[repoName]/groups/[groupId].tsx:13]: I wonder if it might be better to create the heading inside the component rather than as a prop, especially as the edit mode gets added into the mix.

mmeigs (commented):
Looks great, Raymund. Would you like to add the user flow after adding a group now or later on your edit/delete PR? I am okay with this being combined later because there's a lot of bleed-over, but it's up to you!
Also, thank you for InputLabel, InfoTooltip, and FormWrapper. They work like a dream.

-- will look more at your test files tomorrow! Sorry to split my review between days AGAIN!

frontend/src/components/Forms/styles.tsx
rayangler: Agreed! Noticed that too during implementation and it was a little concerning 😆 We should check in with the design team in the upcoming weeks to see if they have any recommendations for styling across different breakpoints

frontend/src/pages/repos/[repoName]/groups/[groupId].tsx
Comment (rayangler) [reply frontend/src/pages/repos/[repoName]/groups/[groupId].tsx:13]: I only separated this into a prop at the time because I didn't want to give the component more responsibility and logic to handle how to show the heading. But you're definitely right that the edit mode logic will probably get in the way anyways. I'll make the change

frontend/tests/unit/GroupForm.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('only allows ungrouped versions to be selected', async () => {
+    const wrapper = render(<GroupForm heading={NEW_GROUP_HEADING_TEXT} repoName="docs" />);
+    const select = wrapper.getByLabelText('Add a Version to This Group');
+    fireEvent.click(select);
+
+    const listbox = wrapper.getByRole('listbox');
+    // Unused versions are those that are not used in a group
+    const expectedUnusedVersions = 1;
```
mmeigs, line 39: Ideally we could make this more responsive and based off of the mockData itself, but I don't think that's worth holding off on this PR to work on.

frontend/src/components/Forms/GroupForm.tsx
rayangler: I don't think there's a difference between `Body` and how `p` is being used here atm! I used `p` as a placeholder during the whole LG Provider error shenanigans but I'll change this to be `Body`, thank you for the reminder!

The `h1` is a little tricky atm because it's a very specific style that I don't think is a preset style in LG's Typography. It's a combination of Typography's `H3` and `Body` and custom font styling. Because of how unique it is, I think it makes sense to not use LG's Typography component here and instead just use our own custom styling

mmeigs: You're totally right! I actually just heard about all this from Allison this morning. 😭 - html h1 is good with me!

frontend/src/components/Forms/GroupForm.tsx
rayangler: Created a custom `FormFooter` component for this where we can pass in an `onCancel` and `onUpdate` (but maybe this should've been called `onSubmit` to make it more generic?) functions as props. Also used LG's constants. Let me know if this looks okay to you!

frontend/tests/unit/GroupForm.test.tsx
rayangler: Good idea! I wrote a small function to handle determining the number for this. It makes the assumption that the data is correct (since working with bad data should be handled by the backend anyways), but hopefully it's flexible enough for our use case?

frontend/tests/unit/GroupForm.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    const inputEl = wrapper.getByPlaceholderText('Name your group') as HTMLInputElement;
+    expect(inputEl.value).toBe('');
+
+    const includedVersionsList = wrapper.getByRole('list');
+    expect(includedVersionsList.childElementCount).toBe(0);
+  });
+
+  it('only allows ungrouped versions to be selected', async () => {
+    const wrapper = render(<GroupForm repoName="docs" />);
```
mmeigs, line 51: nit - can use `testRepoName` here too!

frontend/src/components/Forms/GroupForm.tsx
mmeigs: The FormFooter component looks great. I do prefer onSubmit although that is in conflict with the design.

rayangler: Changed to `onSubmit`. Even though "update" is the term used for the button, it should still be recognized that that button's actually being used to submit the form

frontend/tests/unit/GroupForm.test.tsx
rayangler: Done. Thanks for catching this!

mmeigs (approved):
LGTM! Exciting!

PR #27, opened 2022-08-12
Fix build error from missing packages

rayangler:
### Context

With the current `package-lock.json` file on the `main` branch, `npm run build` would fail locally. This command would not fail for the staging site build because we currently install dependencies via `npm install` instead of `npm ci`, and we don't copy the `package-lock.json` file. This means that the resulting Docker image contains different packages than we would locally.

This PR does the following:

1) Resets the `package-lock.json` file. This was done by deleting the file locally and then running `npm install` with no changes to `package.json`.
2) Updates the `Dockerfile` to copy the `package-lock.json` file and use `npm ci` to install dependencies.
3) The `package-lock.json` update effectively upgrades `gatsby` to v4.20. Since it is a core dependency for us, I decided to explicitly upgrade it in the `package.json` file as well. I also mentioned offline that this version of `gatsby` seems to resolve this issue, so it's worth keeping anyways.

### Open Questions

Open questions for us (not needed to be answered now, but things that came to mind):

1) Why/how did the package dependencies start to differ so much? Not sure if we did anything in particular to cause this, but it is concerning that this difference occurred and resulted in such a big diff over such a short period of time. My investigations haven't yielded a concrete answer, other than maybe this is just the result of upgrading `gatsby` versions.
2) Is anyone able to reproduce the core issue on the `main` branch locally? I feel like a sanity check would be nice to ensure that there isn't something else that's causing this issue on an individual basis instead.


drew-beckmen (approved):
These changes all make sense to me - thanks @rayangler! In terms of your questions, I don't have any concrete answers though there was lots of version wrangling when we were dealing with the LG / React 18 incompatibility issues. I'm in favor of getting this out sooner rather than later so we can avoid these weird diffs in `package-lock.json` between `main` and feature branches!

frontend/Dockerfile
```diff
@@ -2,8 +2,9 @@ FROM node:16-alpine
 ARG GATSBY_API_URL=$GATSBY_API_URL
 WORKDIR /app
 COPY ./package.json .
+COPY ./package-lock.json .
 RUN apk add --no-cache --update python3 make g++
-RUN npm install
+RUN npm ci
 COPY ./src/ ./src
 COPY ./static/ ./static
 COPY ./gatsby-config.ts .
```
drew-beckmen, line 10: One (non-blocking) comment that's unrelated to these changes, but may help reduce the size of our image is to combine the `COPY` statements into a single line. That aligns with Kanopy's suggestion in their[ Docker best practices docs.](https://github.com/10gen/kanopy-docs/blob/master/docs/development/dockerfile_best_practices.md#dont-separate-similar-commands)

mmeigs (approved):
Will definitely keep my eye out for this issue, but I think these changes are good regardless.

frontend/Dockerfile
drew-beckmen: agreed!

rayangler:
> lots of version wrangling when we were dealing with the LG / React 18 incompatibility issues

Right, I forgot that that was bundled into one of the PRs. Good point

PR #28, opened 2022-08-15
DOP-3194: Remove trailing slashes from API, add custom APIRouter to support trailing slashes without 307 redirect

drew-beckmen:
### Ticket

[DOP-3194](https://jira.mongodb.org/browse/DOP-3194)

### Description

This PR resolves an issue where the we received mixed content errors on the frontend despite sending requests via HTTPS. The reason for the error was that our API routes included trailing slashes, so FastAPI (or Starlette, FastAPI's dependency for routing and other web service stuff) redirected our frontend requests to the endpoints with trailing slashes. However, the redirects occurred with HTTP, causing the browser to complain.

After this PR, we will:

- remove trailing slashes from endpoints exposed in the Swagger page
- support requests both with and without trailing slashes using a custom `APIRouter` class. The class automatically adds two decorators to every endpoint, so future API developers won't have to think about this distinction. By supporting both `""` and `"/"` for every endpoint, we also abstract away this distinction from any frontend developers writing code that consumes the API.


backend/api/v1/routes/root.py
```diff
@@ -11,6 +13,6 @@ class ApiDescription(BaseModel):
     description: str
 
 
-@router.get("/", response_model=ApiDescription, tags=["default"])
+@router.get("/metadata", response_model=ApiDescription, tags=["default"])
```
drew-beckmen, line 16: Needed to add an actual route here due to [this known issue](https://github.com/tiangolo/fastapi/issues/510) with empty paths in sub-sub routers.

backend/api/api_router.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Callable
+
+from fastapi import APIRouter as FastAPIRouter
+from fastapi.types import DecoratedCallable
+
+"""
+    Custom APIRouter that supports adds two decorators per endpoint, one
```
rayangler, line 7: (Nit) "supports adding" instead of "supports adds" ?

mmeigs (approved):
LGTM. I personally want to look more into the '/metadata' workaround - more a note to myself to understand what that change entails.

backend/api/v1/routes/root.py
drew-beckmen: Alternatively we could use the `("$")` regex workaround as shown in the GH issue. Open to suggestions...

rayangler: Since we don't *really* need a root route to `/api/v1`, I think keeping this to `/api/v1/metadata` is reasonable to keep? 

Based on the issue comments, seems like using `$` might have implications we'd need to observe, such as on the OpenAPI spec.

backend/api/v1/routes/root.py
```diff
@@ -11,6 +13,6 @@ class ApiDescription(BaseModel):
...
 async def root(settings: Settings = Depends(get_settings)):
```
rayangler, line 17: (Nit) Might make sense to rename this method from `root` to `metadata` if we go with this change

rayangler (approved):
LGTM, but mostly just non-blocking nits. Haven't gotten the chance to test it out locally yet, but changes make sense 😄

backend/api/v1/routes/root.py
drew-beckmen: agreed! will change

backend/api/api_router.py
drew-beckmen: I think I meant to just type "adds", which I feel is better given that the class doesn't _support_ adding two decorators, it actually does the adding of two decorators. Will remove the word "supports"

PR #33, opened 2022-08-26
DOP-3174: Edit and delete functionality for version form page

mmeigs:
### Ticket

[DOP-3174](https://jira.mongodb.org/browse/DOP-3174)

### About

- Added edit and delete functionality to Version Form Page.
- Delete functionality uses the DeleteConfirmationModal component in the correct flow.
- Reorganized Version API functionality into singular util function.
-- If we choose to keep this pattern, we could generalize and leverage this for groups as well


frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -93,13 +62,26 @@ const VersionForm: React.FC<VersionFormProps> = ({ repoName, version }) => {
...
+        (id) => navigate(`${versionsEndpoint}/${id}`),
+        formState
+      );
     } else {
-      console.log('TODO: editing');
-      setDisabled(true);
+      versionAPI(
+        'PUT',
+        `${versionsEndpoint}/${version.id}`,
+        () => window.location.reload(),
```
rayangler, line 75: Just curious: why do we reload the page instead of setting the state to show the non-editable state of the form?

rayangler (approved):
LGTM! 

I was just curious about the reason for a page reload on version update instead of allowing the form to re-render.

frontend/src/components/Forms/VersionForm.tsx
mmeigs: Hm, I think I was so wrapped up in navigate not working for this case that I was stuck in that pattern. That makes sense. Playing around with it now. There is actually an interesting bug happening where both the repo page and the version page doesn't refresh it's state when navigating back and forth... I'm guessing because of the SSR? Refreshing the page does take in the new state. Looking into it now

rayangler: I'm okay with reloading the page for now, especially if it seems like the most reasonable approach with that weird bug. I would just leave a short comment right above that line to explain why we're reloading.

PR #34, opened 2022-08-30
DOP-3216: Add cookies to SSR fetches

rayangler:
### Ticket
DOP-3216

### Notes

* Frontend pages that fetched backend API data were hitting Kanopy's CorpSecure proxy because server-side fetches were not passing cookies. (Client-side fetches pass along cookies just fine.) This change takes the cookies present in the page prop's request headers and passes it along to the backend accordingly, much like how client-side fetches automatically would. This was verified on our staging link.
* No backend logic was changed to accomplish this and the backend API should still be secure. Initiating a request while passing along cookies with a sample JWT to the API staging link resulted in hitting the CorpSecure proxy, as desired and expected.


mmeigs (approved):
LGTM!

PR #35, opened 2022-08-31
DOP-3176: Implement editing and deletion of groups

rayangler:
### Ticket

DOP-3176

### Notes

* Updated the group form to use similar patterns, components, and variables as the version form to keep things more consistent.
* Not implemented: modal appearance when cancelling editing with unsaved changes. The design for this currently shows a mix of two variants (danger icon with green primary button). Will need to confirm with design what the intended design for this is + what the copy should be for groups.
* Editing an existing group will lead to a hard page reload/refresh similar to the version edit functionality introduced in #33. For context, it seems like client-side navigation doesn't cause a re-fetch of SSR'd data ([docs](https://www.gatsbyjs.com/docs/reference/rendering-options/server-side-rendering/#how-it-works)). We'll most likely have to change the group and version forms to use client side fetching instead with skeleton screens (will need to reach out to design about this for intended UX). This change would be beneficial in allowing us to persist any global states in between navigation of pages and re-renders (such as for global toasts), and would help prevent stale data from persisting between client-side navigations.


mmeigs (approved):
Looked over it twice! It's great. LGTM

PR #37, opened 2022-09-02
DOP-3198 - Ensure list of docs repos in side navigation is unique

carooobregon:
### **Ticket**
[DOP-3198](https://jira.mongodb.org/browse/DOP-3198)

### **Notes**
- Fixed it by splitting repo names and making array a Set to eliminate duplicates
- Wondering if there's a problem if the key on `             
            <SideNavItem
            key={repo} /> `
is changed to `repoName` instead of the original `repo` ? From what I see it seems to be behaving normally but just wanted to point out as `repo` includes the whole entitlement name (ex `mongodb/docs-404`) while repoName would just be `docs-404`


frontend/src/components/SideNav.tsx
```diff
@@ -43,12 +50,11 @@ const SideNav: React.FC = () => {
...
       </SideNavItem>
       {entitledRepos?.length > 0 &&
-        entitledRepos.map((repo) => {
-          const repoName: string = repo.split('/')[1];
+        entitledRepos.map((repoName) => {
           return (
             <SideNavItem
               className={cx(navItemStyle(isActive(repoName)))}
-              key={repo}
+              key={repoName}
```
rayangler, line 57: > Wondering if there's a problem if the key on <SideNavItem key={repo} /> is changed to repoName instead of the original repo ?

I think this change should be fine. `repoName` should now be unique anyways

rayangler (commented):
Looks mostly good! Just one nitpick

mmeigs (commented):
LGTM! Just make sure to run `npm run lint:fix` and `npm run format:fix`. Then just make sure the tests are still good to go! :)

frontend/src/components/SideNav.tsx
carooobregon: thanks for clearing it up!

carooobregon:
> LGTM! Just make sure to run `npm run lint:fix` and `npm run format:fix`. Then just make sure the tests are still good to go! :)

thanks for the heads up!

rayangler (approved):
LGTM!

PR #38, opened 2022-09-07
Fix broken async tests

rayangler:
### Problem/Context

Certain tests were emitting warnings and errors while others were being marked as false passes, mostly due to asynchronous operations.

### Solution

The solution was to use async/await on calls to `waitFor`. Without awaiting, `waitFor` will keep trying to run the operations until it timeouts. 

It is assumed that `VersionForm.test.tsx` was more visibly failing than `GroupForm.test.tsx` because there were enough tests in the VersionForm test suite to allow `waitFor` to timeout and emit the error during a different test before the test suite was finished. This would lead to the incorrect test further down the test suite being marked as a fail, despite an `expect` statement from an earlier test failing.

### Notes

Please let me know if anything looks out of place and/or you see tests failing, locally or otherwise!


frontend/tests/unit/VersionForm.test.tsx
```diff
@@ -75,6 +74,11 @@ describe('VersionForm component', () => {
       });
 
       [...textInputs, ...toggles].forEach((input) => {
+        // Stable branch exists, so this toggle should be disabled.
+        if (input.id === 'isStableBranch') {
+          expect(input).toBeDisabled();
+          return;
+        }
```
rayangler, line 81: For some reason, mocking `global.fetch` instead of the `useHapleyApi` hook was not marking the toggle for the stable branch as disabled, even though the mock data already has a stable version. 

Not sure why this was happening exactly, but I went ahead and modified the test to reflect the data more accurately. Please let me know if the previous mocking and behavior were ideal and I could change this back! It's possible that I misunderstood the intended behavior here

frontend/tests/unit/VersionFormErrors.test.tsx
```diff
@@ -40,13 +40,13 @@ describe('VersionForm error states', () => {
...
+    const textInputs = getAllByRole('textbox');
+    fireEvent.change(textInputs[0], { target: { value: usedGitBranchName } });
 
-    const submitBtn = await getByText('Update Document');
-    await fireEvent.click(submitBtn);
+    const submitBtn = getByText('Update Document');
+    fireEvent.click(submitBtn);
 
-    await getByText(
+    getByText(
```
rayangler, line 49: Removed the `await`s here because VS Code was telling me that `await` wasn't actually doing anything here, which is valid

frontend/tests/unit/VersionForm.test.tsx
```diff
@@ -83,7 +87,7 @@ describe('VersionForm component', () => {
   it('creates a new version', async () => {
     const expectedGitBranchName = 'v.2';
     const expectedActiveState = true;
-    const expectedAlias = 'current';
+    const expectedAlias = 'current2';
```
rayangler, line 90: Mocked data already has `current` as an alias. Updating the document would hit the error state and would fail to call `navigate` otherwise. Updated this variable to hit the successful codepath

rayangler (commented):
Leaving annotations for some of the changes that might seem random or questionable out of context

frontend/tests/unit/VersionForm.test.tsx
mmeigs: Oh wow, good catch!

frontend/tests/unit/VersionForm.test.tsx
mmeigs: Great. This must have also been a problem with the global fetch instead of Hapley on my end again. Thank you!

mmeigs (approved):
Thank you for figuring this out. LGTM!

frontend/tests/unit/VersionForm.test.tsx
rayangler: Nah, I don't think this was a `global.fetch` issue! It still worked prior to the mocking change. This is part of the test that didn't have `await` in front of the `waitFor`, so it was causing a false positive/success. Including the `await` revealed that this alias error was being hit, as expected 🙂

mmeigs: Oh, great! I need to start making sure every single test fails first. Thank you

PR #40, opened 2022-09-09
DOP-3205: Validate group label input on the frontend

rayangler:
### Ticket

DOP-3205

### Notes

* Added error handling for group label input.
* Updated `GroupForm` to use the `TextField` component and use text constants, similar to how the `VersionForm` does.


mmeigs (approved):
You're so clean! I tried to find something to nitpick. LGTM!

PR #41, opened 2022-09-12
DOP-3177: Drag and Drop Table

mmeigs:
### Ticket

[DOP-3177](https://jira.mongodb.org/browse/DOP-3177)

### Notes

After attempting to implement drag and drop functionality on top of LG Table component, it was discovered that the LG Table [does not support passing indices down](https://mongodb.slack.com/archives/CFFA5BS79/p1659125886867989) to the Row components. Shifting from the LG Table component to a standard table tag was simple and effective to handle this new functionality. Keeping an eye out for any styles that may not have been ported over from this change.

- Used react-beautiful-dnd library to create custom and modular Table and Cell components. Each particular Row component (VersionRow, GroupRow, and future GroupFormRow(?)) will need a couple wrappers to fit in the flow of this. 
- Full functionality through the API is supported in this ticket, along with the edge case of backend failure which will result in a very quick reset in the order of the rows (with Error Toast in follow-on ticket to be added to explain the reset to the user)
- The drag and drop icon [used in the Figma designs](https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI?node-id=214%3A40559) was created by a past designer. After talks with Allison Mui, we've changed the icon to the "Menu" LG icon.


mmeigs:
Still researching the pervasive unique key error from React. I believe it's an error within the react-beautiful-dnd library itself. Looking into it to hopefully get rid of that.

frontend/src/components/Repo.tsx
```diff
@@ -42,15 +47,39 @@ const getEmptyState = (itemType: string) => {
...
+      return lastState;
+    }
+  };
+};
+
+export const getVersionRow = (data: Version, index: number, repoName: string) => (
+  <VersionRow data={data} index={index} repoName={repoName} />
+);
+export const getGroupRow = (data: Group, index: number, repoName: string) => (
+  <GroupRow data={data} index={index} repoName={repoName} />
```
rayangler, line 73: Adding keys to each `VersionRow` and `GroupRow` should help prevent the key warning/error from happening

frontend/src/components/Table/styles.tsx
```diff
@@ -1,16 +1,24 @@
...
-export const TableRow = styled(Row)`
+export const TableRow = styled('tr')`
   cursor: pointer;
   height: 56px;
   &:hover {
     box-shadow: 0px 0px 4px ${palette.gray.light1};
   }
 `;
+
+export const VersionActiveContainer = styled('div')`
```
rayangler, line 17: (Nit) Since this is used in both `VersionRow` and `GroupRow`, it might be confusing to have this be called "VersionActive". 

Would it make sense to rename this to be something more general or inclusive? Maybe something like `TableCellActiveContainer`?

frontend/src/components/Repo.tsx
```diff
@@ -72,30 +141,44 @@ const Repo: React.FC<{ repoName: string }> = ({ repoName }) => {
       >
         {repoDisplayName(repoName)}
       </H3>
-      <div>
+      <div key={`version-table`}>
```
rayangler, line 144: (Nit) Is this key necessary? If not, we should probably remove unnecessary keys (here and for other elements) to keep things clean/focused

frontend/src/components/Repo.tsx
```diff
@@ -19,6 +21,9 @@ export const VERSION_COLUMN_NAMES = [
 ];
 export const GROUP_COLUMN_NAMES = ['Name', 'Number of Versions'];
 
+export const IsDraggingContextVersions = React.createContext<boolean>(false);
+export const IsDraggingContextGroups = React.createContext<boolean>(false);
```
rayangler, line 25: Wondering if separating these 2 contexts is needed? They both seem to share the same value of `isDragging`, regardless of separation. (i.e. if `isDragging` is set to `true` in the `Repo` component, then both `Table` components will have `isDragging` set to true in their respective contexts)

rayangler (commented):
This works and looks so great! Thank you for working on this! 

Mostly nits and minor suggestions for now as a first pass (plus some questions). If a suggestion doesn't seem reasonable given the full context you have from working on this, feel free to let me know!

frontend/package.json
```diff
@@ -60,11 +60,12 @@
     "@emotion/eslint-plugin": "^11.7.0",
     "@testing-library/dom": "^7.2.1",
     "@testing-library/jest-dom": "^5.4.0",
-    "@testing-library/react": "^10.0.3",
+    "@testing-library/react": "^11.0.0",
```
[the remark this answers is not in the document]
mmeigs: Good catch. This was when I was trying out a different library to test drag and drop and I needed to upgrade, but I believe we can revert now that I dropped that other library. Checking now!

frontend/src/components/Table/GroupRow.tsx
```diff
@@ -1,30 +1,54 @@
...
-    >
-      <Cell>
-        <TableCell>{data.groupLabel}</TableCell>
-      </Cell>
-      <Cell>
-        <TableCell>{data.includedBranches.length}</TableCell>
-      </Cell>
-    </TableRow>
+    <IsDraggingContextGroups.Consumer>
+      {(isDragging: boolean) => (
```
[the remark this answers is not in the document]
mmeigs: You are beyond correct, yes! Can't believe I didn't think that one through farther! Thanks!

frontend/src/components/Repo.tsx
mmeigs: THAT'S IT!! Oh, I love you!! That was driving me MENTAL. Thank you

frontend/src/components/Repo.tsx
mmeigs: Yes, I should go through to make sure all the unnecessary keys are deleted, for sure!!

frontend/src/components/Repo.tsx
mmeigs: Mmm, that's true. I only separated them halfway. Here's the quandary: should I combine them because we'll never have a huge amount of versions and groups for any repo? Or should I completely separate these to give the slightest efficiency to these re-renders? 
- my original thought was because freezing the styles of each cell could get really expensive in re-renders so I was attempting to cut down on at least *which* table needed to freeze styles
- I guess it's if we think that extra effort is worth it...

mmeigs:
Okay! I have done a major re-structure of the code. Not too much functionally has changed, except using React.memo on the inner components of the rows has given us the re-render efficiency of dreams! ie each cell (in the table that is being altered) only re-renders once when the drag starts and once when the drag ends.

Up for discussion:
- The renaming conventions I used (VersionCells, GroupCells, Row, DragDropContainer, etc.)
- I could further reorganize the DragDropContainer to be used once inside the Table component rather than twice in Repo, but I can't decide if that's more readable or less readable. Up for either way!
- I am very torn on if I should keep the rest as-is OR if I should pull the entirety of the state into each table/dragdropcontainer... This could make the code a little cleaner (no more conditionals to see if it's version or group), but the "isLoading" state would either be lost or it would be a little more clunky. Thoughts?
-- This might have been unclear. => I could bring the fetch itself into the Table component and heavily simplify the Repo component.

frontend/package.json
```diff
@@ -54,7 +54,8 @@
     "react-dom": "^17.0.2",
     "react-helmet": "^6.1.0",
     "react-router-dom": "5.2.0",
-    "stream-browserify": "^3.0.0"
+    "stream-browserify": "^3.0.0",
+    "react-beautiful-dnd": "^13.1.1"
```
rayangler, line 58: (Super duper nit) Can we move this to be between `react` and `react-dom` to keep consistent with npm's ordering of packages? I'm surprised the `package-lock` file sorts it out as expected but the `package` file doesn't

frontend/tests/unit/Repo.test.tsx
```diff
@@ -46,3 +51,55 @@ describe('Repo', () => {
     });
   });
 });
+
+describe('Drag and Drop', () => {
+  it('moves a row up', async () => {
+    const versions = getVersions('docs').branches;
+    await render(<Repo repoName="docs" />);
```
rayangler, line 58: (Nit) VS Code is saying that this `await` isn't actually doing anything. Can we remove it? Removing it and re-running the tests seem to pass for me locally (Same thing for lines 68 and 83)

rayangler (commented):
Just a couple of nits, but this is really great! Doesn't seem like every cell from both tables are re-rendering anymore, which is HUGE!

frontend/tests/unit/Repo.test.tsx
mmeigs: Absolutely! And what is the extension you're using?? I would LOVE to have it

mmeigs: Nevermind! Just never noticed those ellipses!!

rayangler:
> The renaming conventions I used (VersionCells, GroupCells, Row, DragDropContainer, etc.)

I think they're all fine!!

> I could further reorganize the DragDropContainer to be used once inside the Table component rather than twice in Repo, but I can't decide if that's more readable or less readable. Up for either way!

I don't have a strong opinion on this, but the benefit of having `DragDropContainer` exclusively in the `Table` component is that it removes the responsibility away from the `Repo` component to remember to use those containers. (i.e. everything related to draggable table rows is strictly only in the table component, except for whatever happens after the drag ends.) I'll leave the final decision up to you on this since I don't think it matters too much at the moment.

> I am very torn on if I should keep the rest as-is OR if I should pull the entirety of the state into each table/dragdropcontainer... This could make the code a little cleaner (no more conditionals to see if it's version or group), but the "isLoading" state would either be lost or it would be a little more clunky. Thoughts?
-- This might have been unclear. => I could bring the fetch itself into the Table component and heavily simplify the Repo component.

I'm fine with keeping the rest as-is. There's already a lot of work done here to add the draggable row functionality to tables. Optimizing it further can come in a future ticket if desired. Also, there's an open concern about keeping this component general enough to accommodate rows/cells for included versions on the group form.

rayangler (approved):
LGTM as-is! If you decide you want to make further changes as part of this PR though, feel free to request a re-review

frontend/tests/unit/Repo.test.tsx
rayangler: Yeah, they're so easy to miss!

rayangler (approved):

PR #42, opened 2022-09-13
DOP-3134: Set up repo data archiving on the backend

rayangler:
### Ticket

DOP-3134

### Notes

The general flow of archiving will be:

1) User performs an update (POST, PUT, DELETE) on versions or groups.
2) User receives the expected response.
3) The *updated* repo + the current time + the current user id (object id from the user's entitlement document) are all saved into a single document in `repos_branches_archive`.

This was implemented using FastAPI's [Background Tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/) feature, which can perform actions in the background after a successful API response.

A documentation (spec) change will be needed to more accurately correspond with the order in which the changes are archived. As outlined in the version management UI spec and its predecessor, the *original* repo document (prior to change) will be archived whenever a change is made. However, this could be inconvenient such that to see who changed the data and what the data was changed to, we'd have to look at 2 different documents. Additionally, this could potentially lead to gaps in changes if someone were to edit data directly through the db.


mmeigs (approved):
Looks lovely! LGTM

PR #43, opened 2022-09-13
DOP-3226: Implement Cancellation Modal

mmeigs:
### Ticket

[DOP-3226](https://jira.mongodb.org/browse/DOP-3226)

### Notes

The group and version forms should display a modal that asks the user if they want to discard any changes they've made when cancelling.

- Confirmed [new design with Allison Mui](https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI?node-id=817%3A22533) to create a LG-usable confirmation modal for this warning on discarding changes.
- As shown by the Figma designs:
-- When the user hits `Cancel`, the user is taken back to the current "changes" that they were about to discard.
-- When the user clicks `Leave without saving`, the user is taken all the way back to the repo page with the list of versions for that repo, and none of their changes are saved.


rayangler (commented):
Mostly looks good to me, but just wanted to ask about a behavior I was running into locally

frontend/src/hooks/use-form-state.tsx
```diff
@@ -23,7 +23,8 @@ function formReducer(state: Version, action: FormAction) {
       const { property, value } = payload;
       return {
         ...state,
-        [property]: value
+        [property]: value,
+        hasBeenEdited: true
```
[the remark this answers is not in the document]
mmeigs: Ohhh deep. Okay. I found a way to make this happen. Just fighting with typescript now.

mmeigs:
@rayangler I've now modified use-form-state to check updates before setting the `hasBeenEdited` property. Let me know what you think

frontend/src/hooks/use-form-state.tsx
```diff
@@ -16,38 +19,52 @@ type FormAction =
...
+      case 'UPDATE':
+        const { property, value } = payload;
+        const prop = property as keyof typeof initialState;
+
+        let hasBeenEdited = initialState[prop] !== value;
+        if (property === URL_ALIASES) {
+          const initialAliases = initialState[prop] as string[];
+          const currentAliases = value as string[];
+          hasBeenEdited = !(initialAliases.length === currentAliases.length);
+          !initialAliases.every((el: string, i: number) => el === currentAliases[i]);
```
rayangler, line 35: Seems like the return value of this isn't being used for anything

frontend/src/hooks/use-form-state.tsx
```diff
@@ -16,38 +19,52 @@ type FormAction =
...
-  }
+function createFormReducer(initialState: Version) {
+  return function formReducer(state: Version, action: FormAction) {
+    const { type, payload } = action;
+    switch (type) {
+      case 'UPDATE':
+        const { property, value } = payload;
+        const prop = property as keyof typeof initialState;
+
+        let hasBeenEdited = initialState[prop] !== value;
```
rayangler, line 30: I think the value of this would be reset every time a different field is updated. For example: 

1) edit git branch name
2) edit version selector label
3) edit version selector label back to its original value
4) cancel

The form cancels without the confirmation modal as if the git branch name has not been changed

frontend/src/hooks/use-form-state.tsx
mmeigs: Oh yeah... this is so interesting to me! It linted it that way... I just changed it and linted again and it looks like it's better now. Still unsure of what it thought I was doing or if that actually works as a logical and... probably not.

frontend/src/hooks/use-form-state.tsx
mmeigs: ohp yeah. You're right. Okay. I really just did not want to have to do a huge comparison when we hit cancel. But I guess that's the way I'm going now.

mmeigs:
I've moved back to my original thought of doing one full comparison of each property when hitting Cancel. Sorry that I've been all over the place with this PR! Let me know if this feels like it needs another iteration

rayangler (approved):
LGTM! Thanks for working on this

PR #45, opened 2022-09-16
DOP-3249: Upgrade LG Table package

mmeigs:
### Ticket

[DOP-3249](https://jira.mongodb.org/browse/DOP-3249)

### Notes

Necessary security upgrade for LG Table elements/package to v6.0.1
Reasoning: [spoken of in Slack post](https://mongodb.slack.com/archives/CFFA5BS79/p1663340497141869)

- Also, investigated and found only "breaking" change 
-- manually reinserted style rule into Hapley for vertical alignment


rayangler (approved):
LGTM!

Wanted to note that the new version of LG's `Table` has a peer dependency of the LG Provider component v2.3.3. We're currently on v2.3.2, but the package-lock seemed to have updated to v2.3.3 automatically (but the package files has not). Doesn't seem like a problem atm but we should be mindful of this in case any weirdness with our dependencies happens in the near future

mmeigs:
> 

Great catch!!

PR #25, opened 2022-08-11
DOP-3187: Create hook for fetching from Hapley API, add global error toast

drew-beckmen:
### Ticket

[DOP-3187](https://jira.mongodb.org/browse/DOP-3187)

### Description

This PR includes a wrapper around the default LG warning toast that sources data from an `ErrorContext`. Using the `setError` from `ErrorContext`, we'll be able to set errors and reuse the same error component across the application. By default, the global error toast disappears after 5 seconds on the screen.

I've also refactored the `fetch` call out of the `SideNav` and `Repo` component into a custom hook. The hook allows us to reuse code in the `fetch` process, namely around error handling. I've leveraged TypeScript generics so that each component that leverages the hook can pass an interface that describes the structure of the API response.

The hook implements the following behavior when handling error responses for GET requests:

- 404 -> redirect to the 404 page
- 401 unauthorized -> redirect to homepage and display error toast with unauthorized message
- all other non-200 responses -> display error toast with message on the user's current page

### Screenshot

Screen Shot 2022-08-11 at 9 59 16 AM

PR #47, opened 2022-09-16
DOP-3206: Implement global toast

rayangler:
### Ticket

DOP-3206

### Notes

* Generalizes the implementation of the error toast in #25 in the hopes of a generally reusable toast.
* Adds toast triggers to version and group forms. The toast for edits will most likely not run while we continue to reload the page.
* Uses version 1.0.1 of the LG Toast component instead of using latest (2.0.0). Using 2.0.0 leads to incorrect rendering of CSS for the Toast component (the close button ends up in the incorrect position and color). Attempting to upgrade LG's emotion package fixes Toast rendering but results in rendering issues for LG's Side Nav's button. In the interest of time and to avoid any confusing dependency entanglements and perhaps out-of-scope changes, I decided to stick with the second-most latest version in the hopes that we can sort out any dependency issues later.


frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -200,7 +218,12 @@ const GroupForm: React.FC<GroupFormProps> = ({ repoName, group }) => {
 
   const handleModalConfirm = useCallback(() => {
     if (modalType === 'delete') {
-      repoItemAPI('DELETE', `${groupsEndpoint}/${group?.id}`, () => navigate(backButtonLink));
+      // Deletion should only be possible if the group exists.
```
mmeigs, line 221: I understand the logic of this, but I believe the only way that the code will get to handleModalConfirm at all is if a group/version exists. The edit/delete buttons have to be active, which means that it has to be a group/version that already has been created.

frontend/src/context/toast-context.tsx
```diff
@@ -0,0 +1,68 @@
...
+  triggerToast: () => {}
+});
+
+const ToastContextProvider: React.FC<ProviderProps> = ({ children }) => {
+  const [toastProps, setToastProps] = useState<ToastProps | null>(null);
+  const toastOpen = !!toastProps;
+
+  const triggerToast = useCallback((variant: Variant, message: string) => {
+    setToastProps({ variant, message });
+  }, []);
```
rayangler, line 32: Alternatively, I thought about turning this into an object with functions that correspond to the variants of toast we expect to use (success and warning?). This way, we wouldn't need to pass in the variant for each trigger albeit we'd need to write repetitive functions and calls to `setToastProps`.

Decided to stick with this more generalized function call for now but curious if there's a preference at the moment. (The ideal usage might become more apparent over time as more triggers of the toast are needed)

frontend/src/components/Forms/GroupForm.tsx
mmeigs: I put the optional chaining to shut typescript up I think lol

frontend/tests/unit/Toast.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    expect(wrapper.queryByText(TOAST_MESSAGE)).toBeFalsy();
+  });
+
+  it('closes automatically after 5 seconds', async () => {
+    jest.useFakeTimers();
+
+    // Allows callback in setTimeout to run after being clicked
+    const user = userEvent.setup({
+      advanceTimers: () => jest.runOnlyPendingTimers()
+    });
```
rayangler, line 56: TIL that using fake timers does not work well with RTL's `waitFor` function (it seems to mess with how `waitFor`'s time works too?). `userEvent` seems to provide a workaround that can automatically advance fake timers after an event

(Somewhat unrelatedly, [RTL's docs](https://testing-library.com/docs/dom-testing-library/api-events/) suggests we should probably use `userEvent` most of the time, even if their examples do use `fireEvent` 😅 )

frontend/src/constants.ts
```diff
@@ -60,7 +60,7 @@ export const MODAL_CONSTANTS = {
   },
   delete: {
     title: (item: string, repo?: string) =>
-      `Are you sure you want to delete ${item}${repo && ` in ${repo}`}?`,
+      `Are you sure you want to delete ${item}${!!repo ? ` in ${repo}` : ''}?`,
```
rayangler, line 63: Didn't catch this during review, but this was resulting in "undefined" appearing as part of the text for the group form when deleting something. The change is out of scope but figured it was small enough to include here

frontend/src/context/toast-context.tsx
mmeigs: Hmm, I'm really half and half on this. Thought about it for a couple minutes and couldn't make up my mind. Let's leave it like this! Like you said, easy change in the future!

frontend/tests/unit/Toast.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    await userEvent.click(closeButton);
+    expect(wrapper.queryByText(TOAST_MESSAGE)).toBeFalsy();
+  });
+
+  it('closes automatically after 5 seconds', async () => {
+    jest.useFakeTimers();
+
+    // Allows callback in setTimeout to run after being clicked
+    const user = userEvent.setup({
+      advanceTimers: () => jest.runOnlyPendingTimers()
```
mmeigs, line 55: I am fascinated by this!! No comment, just never seen it!

mmeigs (commented):
LGTM! Just really one comment on something that might not be needed in the forms!

frontend/src/components/Forms/GroupForm.tsx
rayangler: I agree that the check for an existing version/group is functionally redundant. The only reason I included it was because optional chaining would result in `group?.groupLabel` being type `string | undefined`. We could potentially get around this by just making `itemName` optional when constructing our toast message, but this would contradict/violate the logical expectation of `groupLabel` being a required part of the message in the toast.

It seems like a matter of functionally redundant logic vs. logically contradicting typing and I'm honestly not sure what the best thing to do here is. I'll spend more time thinking about this on Monday

mmeigs (approved):
LGTM :)

PR #48, opened 2022-09-19
DOP-3228 -  Responsive screens for form pages

carooobregon:
**Ticket**
[DOP-3228](https://jira.mongodb.org/browse/DOP-3228)

**Description**
Implemented designs from [Figma](https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI?node-id=214%3A37309) to make Forms page responsive.
When screensize is lower than 739,  to prevent content being hidden from the small viewport window, margin is calculated using the viewportsize and 370 which is the maximum width the content gets to be.


mmeigs (commented):
Looks great already! I'm seeing three things that might be quick changes to make it better.

- Could we wrap the Update Document and Cancel buttons sooner so that "Update" and "Document" will never wrap in the button? 
- The space between the Update Document buttons and the Cancel button when they are on top of each other is great! Can we add that same margin to the Edit and Cancel buttons?
- There's a slight bug if gitBranchNames are really really long. The figma designs (https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI?node-id=973%3A24430) have them cut off and turned into ellipses which is a great way to deal with it. Maybe try to do this, but if that's too intricate, you could also just make sure that the screen is scrollable! Because if you add a new version and give it a really long name, then go to a group and add that version, and then make the screen small - you can't scroll over to see the rest of the table!

carooobregon:
> Looks great already! I'm seeing three things that might be quick changes to make it better.
> 
> * Could we wrap the Update Document and Cancel buttons sooner so that "Update" and "Document" will never wrap in the button?
> * The space between the Update Document buttons and the Cancel button when they are on top of each other is great! Can we add that same margin to the Edit and Cancel buttons?
> * There's a slight bug if gitBranchNames are really really long. The figma designs (https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI?node-id=973%3A24430) have them cut off and turned into ellipses which is a great way to deal with it. Maybe try to do this, but if that's too intricate, you could also just make sure that the screen is scrollable! Because if you add a new version and give it a really long name, then go to a group and add that version, and then make the screen small - you can't scroll over to see the rest of the table!

Thanks for the comments @mmeigs ! Just updated the PR with the requested changes

mmeigs (commented):
:) So pretty! Just a couple comments to combine code and an odd style that I think is left-behind! Also, I just found a couple styling errors that I did in my last PR that I need to fix, so this was great to dive into.

frontend/src/components/Forms/components/FormHeader.tsx
```diff
@@ -33,6 +34,22 @@ const requiredStyle = LeafyCSS`
   margin: 0;
 `;
 
+const StyledButton = styled(Button)`
```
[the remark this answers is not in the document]
carooobregon: Good catch I hadn't seen that! I changed it so that it follows a similar pattern like the other buttons where the width is specified by the container and just use the flex box properties to adjust the button size

frontend/src/components/Forms/components/FormFooter.tsx
```diff
@@ -13,16 +13,27 @@ const Container = styled('div')`
...
   column-gap: 12px;
   padding: 40px ${FORM_PADDING_LEFT_RIGHT};
+  @media only screen and (max-width: 810px) {
+    flex-wrap: wrap;
+    flex-direction: row;
+    row-gap: 16px;
+  }
+`;
+
+const StyledButton = styled(Button)`
```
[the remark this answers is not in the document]
carooobregon: Great idea! Added the similar properties to `styles.tsx` and used `Styled` in each component for unique properties

frontend/src/components/Table/IncludedVersionsCells.tsx
```diff
@@ -12,9 +14,24 @@ interface IncludedVersionCellsProps {
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: ${offset ? 200 - offset : 200}px;
+`;
+
 // memoized cells to cut down on renders
 const IncludedVersionCells = React.memo(
   ({ isDragging, data, disabled, removeVersion, ...rest }: IncludedVersionCellsProps) => {
+    const viewportSize = useViewportSize();
+    var offset;
```
rayangler, line 28: (Nit) Can we use `let` here? We should avoid using `var` unless necessary

frontend/src/components/Forms/components/FormWrapper.tsx
```diff
@@ -3,23 +3,37 @@ import { css, Global } from '@emotion/react';
...
+  }
 `;
 
 const FormWrapper: React.FC<FormWrapperProps> = ({ children }) => {
+  // In case viewportSize is null, margin 0 is used as fallback
+  const viewportSize = useViewportSize()?.width || 370;
+
   return (
-    <Wrapper>
+    <div className={cx(wrapper(viewportSize))}>
```
rayangler, line 36: (No change needed, just curious) Why the switch from styled component to LG css?

frontend/src/components/Forms/components/FormWrapper.tsx
```diff
@@ -3,23 +3,37 @@ import { css, Global } from '@emotion/react';
...
+const wrapper = (viewportSize: number) => LeafyCSS`
   display: flex;
   flex-direction: column;
   background-color: ${palette.white};
   box-shadow: 0px 4px 10px -4px rgba(0, 30, 43, 0.3);
   margin: 56px 184px;
   min-height: 772px;
+  min-width: fit-content;
+
+  @media only screen and (max-width: 738px) {
```
rayangler, line 22: Should the breakpoint here be 768px? The designs on Figma show 768px, so just wanted to make sure

frontend/src/components/Table/IncludedVersionsCells.tsx
```diff
@@ -12,9 +14,24 @@ interface IncludedVersionCellsProps {
...
+    if (viewportSize && viewportSize?.width < 890) {
+      offset = 890 - viewportSize?.width < 100 ? 890 - viewportSize?.width : 100;
```
rayangler, line 30: (Perhaps nit) To avoid needing to write `890 - viewportSize.width` a couple of times on the same line, we could save this as a constant. 

Also, since we do the check for `viewportSize` in the if statement, we don't need the optional chaining operator since we're sure `viewportSize` exists

frontend/src/components/Forms/components/FormWrapper.tsx
carooobregon: When sending props to styled LG css, typescript complains we're not specifying what props we're sending! I tried using an interface but it still gives me the following error

`'Wrapper' cannot be used as a JSX component.
  Its return type 'StyledComponent<{ theme?: Theme | undefined; as?: ElementType<any> | undefined; }, DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, {}>' is not a valid JSX element.
`

rayangler: Ohh, gotcha! Seems like using an interface works for me locally:

```
interface WrapperProps = {
  viewportSize: number;
};

const Wrapper = styled('div')<WrapperProps>`
  // css here
`;
```

But that being said, I'm in favor of keeping LG css here for simplicity!

mmeigs (approved):
LGTM!

frontend/src/components/Forms/components/FormWrapper.tsx
```diff
@@ -3,23 +3,37 @@ import { css, Global } from '@emotion/react';
 import styled from '@emotion/styled';
```
rayangler, line 3: (Nit) we should remove this, if unused now

frontend/src/components/Forms/components/FormFooter.tsx
```diff
@@ -1,29 +1,33 @@
...
+    <StyledButton variant={Variant.Primary} size={Size.Large} onClick={onSubmit}>
       Update Document
-    </Button>
-    <Button size={Size.Large} onClick={onCancel}>
+    </StyledButton>
+    <StyledButton size={Size.Large} onClick={onCancel}>
       Cancel
-    </Button>
-  </Container>
+    </StyledButton>
```
rayangler, line 29: Should we replace `StyledButton` here with the new `FormButton` styled component you exported in `styles.tsx`?

rayangler (commented):
Just wanted to flag two more nits to help clean up the code a bit, but looks fine overall

frontend/src/components/Forms/components/FormFooter.tsx
```diff
@@ -1,29 +1,38 @@
...
+
 const FormFooter: React.FC<FormFooterProps> = ({ onCancel, onSubmit }) => (
-  <Container>
-    <Button variant={Variant.Primary} size={Size.Large} onClick={onSubmit}>
+  <FormFooterButtonContainer>
+    <StyledButton variant={Variant.Primary} size={Size.Large} onClick={onSubmit}>
       Update Document
-    </Button>
-    <Button size={Size.Large} onClick={onCancel}>
+    </StyledButton>
```
rayangler, line 31: `FormButton` can be used here and `StyledButton` can be removed, unless there's a reason why we need `StyledButton`.

rayangler (approved):

PR #49, opened 2022-09-19
Fix table dependency installation

rayangler:
### Context

Fixes the last 2 build errors on the `main` branch. The build was failing at the installation step when testing the frontend.

### Changes

This PR reverts the changes made in #45, re-installs `6.0.1` of the LG `Table` component, and re-adds the css changes made for vertical alignment. It's unclear what exactly in the installation was causing this, but it might be worth noting differences in `package-lock` changes.

`npm install` was also changed to `npm ci` as part of the Drone pipeline for frontend testing. This was mostly done during debugging, but was kept around to keep things consistent with the GitHub Actions CI's command. There should be no change/difference in this case.

Please feel free to test out locally to ensure that the table component still acts and looks the same.


mmeigs (approved):
Ahh, interesting to see the package-lock. Definitely somehow deleted too many packages... LGTM

PR #50, opened 2022-09-23
DOP-3229: replaced SSR with client-side fetches for Form elements

mmeigs:
### Ticket

[DOP-3229](https://jira.mongodb.org/browse/DOP-3229)

### Notes

It was discovered that Gatsby's implementation of SSR can lead to stale data populating the page on client-side navigation ([Gatsby docs](https://www.gatsbyjs.com/docs/reference/rendering-options/server-side-rendering/#how-it-works)). To ensure that users are getting the most up-to-date data, we switched to client-side navigation for the Form pages.

Many things were changed in this PR.
- Created a util function called `use-hapley-api-swr.ts` that wraps useSWR up in a clean and usable way for both forms.
- Implemented this for both Forms inside a container called FormContainer (more on this later)
- With this switch, I could now change our saving on 'edit' to be a switch of state rather than a page refresh. HOWEVER...
- Therefore, needed to change both forms semi-dramatically to account for many, many pieces of information that then become stale (ie usedValues for validation that are now incorrect, and also dependence on the prop of the Version or Group, because that value has now changed)
- I fixed a styling bug (of my own making) from the creation of the tables
- I added a scroll on the Repo page for now, because I realized when values are long, there is no way to see the rest of the table
- I added and fixed tests

**More on the FormContainer component**:  I originally implemented useSWR directly in both Form components. However, the tests were absolutely buggin' out - "Too many re-renders" errors even though this absolutely never happens in true practice. This is a personal failing... or maybe a problem with testing useSWR overall. Not many resources out there to go on.

To find a solution, I realized I could pull this useSWR functionality outside of the component, which also led to an easier update overall for the data/error/loading states. It gave a much easier testing experience. (I do think the future testing through Cypress can allow us to fully test the entire flow)

I am very up for ideas on this last point. I still have [the original branch](https://github.com/mongodb/docs-hapley/tree/DOP-3229) left how it was (working well, but tests failing) if you are ever interested.

Finally: let me know if you have any questions!


frontend/src/hooks/use-form-state.tsx
```diff
@@ -69,7 +80,16 @@ const useFormState = (initialState: Version | undefined): any => {
     });
   };
 
-  return [formState, dispatchReset, dispatchUpdateField];
+  const dispatchSetLoadedState = (version: Version) => {
+    dispatch({
+      type: 'RESET',
```
rayangler, line 85: Should the type here be `SET_LOADED_STATE` instead of `RESET`?

frontend/src/hooks/use-form-state.tsx
```diff
@@ -14,11 +14,22 @@ type FormAction =
...
     };
 
 function formReducer(state: Version, action: FormAction) {
   const { type, payload } = action;
   switch (type) {
+    case 'SET_LOADED_STATE':
+      return {
+        ...state,
+        ...payload.value
+      };
```
rayangler, line 32: What is the difference between this action type and the `RESET` action type? They seem to have the same logic

frontend/src/components/Forms/FormContainer.tsx
```diff
@@ -0,0 +1,23 @@
...
+import VersionForm from './VersionForm';
+
+type FormContainerProps = {
+  repoName: string;
+  itemId: string;
+  itemType: 'version' | 'group';
+};
+
+const FormContainer = ({ repoName, itemId, itemType }: FormContainerProps) => {
+  const groupEndpoint = `/repos/${repoName}/${itemType}s/${itemId}`;
```
rayangler, line 13: Since this is also being used for versions, we should consider renaming this variable to keep it generalized

frontend/tests/unit/GroupForm.test.tsx
```diff
@@ -156,11 +159,13 @@ describe('GroupForm', () => {
     expect(deleteButton).toBeTruthy();
   });
 
-  it('allows existing group information to be edited', () => {
+  it('allows existing group information to be edited', async () => {
```
rayangler, line 162: Doesn't seem like we're using an `await` statement within this test. Is this `async` doing anything? Should we remove it?

mmeigs: You're right, I forgot to remove!

frontend/src/components/Forms/FormContainer.tsx
mmeigs: Yes! Thanks

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -173,50 +205,50 @@ const GroupForm: React.FC<GroupFormProps> = ({ repoName, group }) => {
...
+        updateGroups(editedGroup);
+        setDisabled(true);
+        setGroupData(editedGroup);
+      });
     }
   }, [groups, groupLabel, includedVersions]);
 
   const resetFormValues = useCallback(() => {
-    setGroupLabel(defaultGroupLabel);
-    setIncludedVersions(defaultIncludedVersions);
```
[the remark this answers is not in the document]
mmeigs: I was having an issue earlier with them not changing correctly with the data, but now they seem to be correct... Maybe that was from when the useSWR was in the Form component itself. I think you're right now that the prop would be updating, it would work. Let me change it back, I did like your defaults.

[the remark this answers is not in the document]
mmeigs: Also, now we don't need the reset functionality anymore! This helped me realize!

frontend/src/components/Layout.tsx
```diff
@@ -5,6 +5,7 @@ import SideNav from './SideNav';
 import SiteMetadata from './SiteMetadata';
 import '../styles/global.css';
 import { ToastContextProvider } from '../context/toast-context';
+import { css, cx } from '@leafygreen-ui/emotion';
```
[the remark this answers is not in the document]
mmeigs: Deleted! That was from an experiment and unused. Thanks for catching!

frontend/src/hooks/use-form-state.tsx
mmeigs: You're totally right... and I never even used SET_LOADED_STATE... wow. Sorry about that. Gone!

frontend/src/hooks/use-form-state.tsx
mmeigs: How about I change RESET to SET_STATE or something similar, so that it works for both? That will help keep it general enough to use for the reset functionality and also the updating of the entire state when the fetch comes through

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -108,21 +142,26 @@ const VersionForm: React.FC<VersionFormProps> = ({ repoName, version }) => {
...
-  const [disabled, setDisabled] = useState<boolean>(!isNew);
-
-  const [formState, dispatchReset, dispatchUpdateField] = useFormState(version);
-  const [errorState, dispatchError, clearAllErrors, clearUnsavedAlias] = useErrorState();
-
-  const [addAliasView, setAddAliasView] = useState<boolean>(false);
-  const [openConfirmationModal, setOpenConfirmationModal] = useState<boolean>(false);
-  const [modalType, setModalType] = useState<'delete' | 'edit'>('delete');
+  const initialUsedVersions: Version[] = useHapleyApi(`/repos/${repoName}/versions/`)?.branches;
+  const [usedVersions, setUsedVersions] = useState<Version[]>([]);
```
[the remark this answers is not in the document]
mmeigs: Yes, much better

frontend/src/components/Forms/GroupForm.tsx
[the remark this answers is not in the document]
rayangler: Ah, okay. If they're not behaving/updating as expected, I'm okay with not keeping them. It was just unexpected at first glance reading!

mmeigs:
Fixed the test warnings and implemented feedback. I re-added the default values in the group form, and it seems to be good. Because of the hooks, I don't believe it's totally needed, but I do think it's clearer code so I like it. Let me know if there's anything else!

rayangler (approved):
LGTM! Just one nit for consistency

frontend/src/components/Forms/FormContainer.tsx
```diff
@@ -0,0 +1,23 @@
...
+type FormContainerProps = {
+  repoName: string;
+  itemId: string;
+  itemType: 'version' | 'group';
+};
+
+const FormContainer = ({ repoName, itemId, itemType }: FormContainerProps) => {
+  const endpoint = `/repos/${repoName}/${itemType}s/${itemId}`;
+
+  let Form = itemType === 'version' ? VersionForm : GroupForm;
```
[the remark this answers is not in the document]
mmeigs: Absolutely! Left behind from a prior idea.

PR #51, opened 2022-09-23
DOP-3233 - Text input cleanups

carooobregon:
**Ticket:**
[DOP-3233](https://jira.mongodb.org/browse/DOP-3233)

**Comments:**
- For urlSlug and gitBranch, we detect the special character when it's already submitted as we are using the previously implemented logic of `validateEntries` and using `errorState`
- For url aliases it's a little trickier since there can be several inputs and can lead to user confusion so I added the error handling inside of the `Alias` component.

**Acceptance criteria**
- Verify it's error catching the following fields in Version Form
- Git branch should follow [these branch naming rules](https://stackoverflow.com/a/12093994)
- URLSlug and Alias should allow alphanumerics, periods, hyphens and slashes as discussed on Slack. Anything outside of those characters should return an error


carooobregon:
converting back to draft as i'll be adding special character handling too!

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -143,14 +149,15 @@ const VersionForm: React.FC<VersionFormProps> = ({ repoName, version }) => {
   );
 
   const validateEntries = useCallback(() => {
-    const errorsFound = [];
+    const errorsFound: Function[] = [];
```
mmeigs, line 152: love this, thank you

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -161,11 +168,13 @@ const VersionForm: React.FC<VersionFormProps> = ({ repoName, version }) => {
...
           URL_SLUG,
           `This urlSlug value does not match the gitBranchName or an entry in the urlAliases array.`
-        )
+        ),
+      [`INVALID_URLSLUG`]: () => dispatchError(URL_SLUG, ERRORS.UNEXPECTED_CHAR)
     };
-
     if (!formState[GIT_BRANCH_NAME]) {
       errorsFound.push(errorDispatches[`NO-${GIT_BRANCH_NAME}`]);
+    } else if (!validBranch(formState.gitBranchName)) {
```
mmeigs, line 176: Just to keep consistent, could we use the constant? ie `formState[GIT_BRANCH_NAME]`

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -161,11 +168,13 @@ const VersionForm: React.FC<VersionFormProps> = ({ repoName, version }) => {
         dispatchError(
           URL_SLUG,
           `This urlSlug value does not match the gitBranchName or an entry in the urlAliases array.`
-        )
+        ),
+      [`INVALID_URLSLUG`]: () => dispatchError(URL_SLUG, ERRORS.UNEXPECTED_CHAR)
     };
-
```
mmeigs, line 166: (nitty-nit-nit) Okay, this is fully up for debate, but I find a certain amount of blank lines to help code readability. In this function, there used to be two blank lines (to break up the errorDispatches dictionary from the building of the errors and then from the implementation of firing the error dispatches) that imho help the eye take in the purpose of these blocks of code. How do you feel?

frontend/src/components/Forms/components/AliasInput.tsx
```diff
@@ -44,11 +46,21 @@ const AliasInput: React.FC<AliasInputProps> = ({
...
   const [alias, setAlias] = useState<string>(aliasToEdit || '');
   const [usedAliasError, setUsedAliasError] = useState<boolean>(false);
+  const [invalidCharError, setInvalidCharError] = useState<boolean>(false);
 
   useEffect(() => {
     setUsedAliasError(usedAliases.has(alias) || urlAliasesMinusAliasToEdit.includes(alias));
+    setInvalidCharError(!validChars(alias));
   }, [alias]);
 
+  let aliasError = invalidCharError
```
mmeigs, line 56: This can be a const! Also, my old eyes took a second to read through this - I don't think I'd ever seen a chained ternary. Fun!

mmeigs (commented):
Looks great, Caro! I've added a few small comments. Also, I realized while playing around that there's an edge case we didn't think of. Right now it's possible to save a blank alias! Do you feel comfortable adding that into the scope of this PR?

frontend/src/utils/character-validation.ts
```diff
@@ -0,0 +1,10 @@
+export function validBranch(branch: string): boolean {
+  const bad = /(^|[/.])([/.]|$)|^@$|@{|[\x00-\x20\x7f~^:?*[\\]|\.lock(\/|$)/;
```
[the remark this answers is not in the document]
carooobregon: hahaha im sorry very true!

frontend/src/components/Forms/VersionForm.tsx
carooobregon: yeah def it makes more sense!

frontend/src/components/Forms/VersionForm.tsx
carooobregon: i fully agree with you but the linter doesnt! 😬 is there a way to bypass it or should we just keep it like this?

frontend/src/components/Forms/components/AliasInput.tsx
carooobregon: haha it seems like the easiest/ tidiest way to do it instead of all the ifs! lmk if you think i should break it down to if statements instead, and very true, just changed it to const!

carooobregon:
@mmeigs thanks for the comments! I just did the changes and added corresponding tests. Good catch on the empty alias edge case! I hadn't seen that. I added error handling for that as well so that when it's empty the save button is disabled.

frontend/src/components/Forms/components/AliasInput.tsx
mmeigs: No, it looks good! :)

mmeigs (approved):
LGTM!

carooobregon:
Thanks @mmeigs ! Waiting for @allisonmui9 's response regarding the copy before merging :-)

carooobregon:
oh actually she responded and said that copy works, merging now! thanks to you both :-)

PR #52, opened 2022-09-27
DOP-3253 - Responsiveness for repo pages

carooobregon:
**Ticket:**
[DOP-3253](https://jira.mongodb.org/browse/DOP-3253)

** Figma **
[figma](https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI)
**Comments:**
Added corresponding CSS to Repo pages to make them responsive.
Moved the `overflow: scroll` property to `Layout` which encompasses SideNav and ContentContainer, so had to add styling to those two. Added an extra div to the SideNav (sorry! 😬 ) since it needed a wrapper that was set as relative and SideNav set as fixed. Open to suggestions but played around with a lot of different options and this was the best one that came out!


mmeigs (commented):
Looks very smooth. I'm fine with the addition of the div if it's what we needed to do! 

Two tiny things:
- The mobile topNav is now not shrinking with the screen width. I'm wondering if you can find a way to keep it responsive to the screen so that the user can always hit the burger menu icon without scrolling!
- Nit: I've noticed that the cells become a bit crowded when on the smallest width. I think we could probably give the TableCell styled component some padding-right? Maybe 15px or something so that there's some breathing room at its smallest. Let me know what you think is best!

frontend/src/components/SideNav.tsx
```diff
@@ -47,27 +49,35 @@ const SideNav: React.FC = () => {
...
-            </SideNavItem>
-          );
-        })}
-    </LGSideNav>
+    <div
+      className={cx(LeafyCSS`
+      position: relative;
+      width: 300px;
+      
+    `)}
```
rayangler, line 57: As is, this div seems to continue to take up space on the page, even when the side nav is collapsed:

Screen Shot 2022-09-30 at 10 37 40 AM

I'm not sure what you did/didn't try yet or if this is a good idea, but curious if there's a way to have the horizontal scroll container limited to the actual content rather than the whole page to avoid this issue

PR #56, opened 2022-10-03
[New] DOP-3253 - Responsiveness for Repo Pages

carooobregon:
Ticket:
[DOP-3253](https://jira.mongodb.org/browse/DOP-3253)

** Figma **
[Figma](https://www.figma.com/file/oJ1Cf4M15qOu1CbU70AMrd/Version-Management-UI)

**Comments:**
Added corresponding CSS to Repo pages to make them responsive
- Green buttons stay as wide as the table
- Added margin to `ContentContainer` to add margin between table and end of window when screen is small


frontend/src/components/ContentContainer.tsx
```diff
@@ -8,6 +8,10 @@ const ContentContainer: React.FC<{ children: React.ReactNode }> = ({ children })
       css={css`
         overflow-x: scroll;
         overflow-y: auto;
+        display: flex;
+        @media only screen and (max-width: 768px) {
+          margin-right: 20px !important;
```
carooobregon, line 13: Chose 20px as an arbitrary number since that's what looked best in my opinion but open to suggestions.

mmeigs: I think maybe 72px just to equal out the margin on the left side?

mmeigs (commented):
I think you've cracked the hard part!! The only thing that threw me off was how the margin became white space covering part of the screen. 
I think if you take away the margin-right important rule, and then you add margin-right to each of the RepoItemHeading "HeadingContainer" divs and also the same margin-right to the main Table div it'll work without that blank space!

carooobregon:
> I think you've cracked the hard part!! The only thing that threw me off was how the margin became white space covering part of the screen. I think if you take away the margin-right important rule, and then you add margin-right to each of the RepoItemHeading "HeadingContainer" divs and also the same margin-right to the main Table div it'll work without that blank space!

great idea that worked perfectly!! thanks for the suggestion of setting as 72px margin to equal the left too :))

mmeigs (approved):
LGTM

PR #59, opened 2022-10-04
DOP-3254

mmeigs:
### Ticket

[DOP-3254](https://jira.mongodb.org/browse/DOP-3254)

### Notes

When accessing version and group data for mongocli docs on hapley, the backend returns an error, due to validation errors with aliases. 

Versions often include an old/deprecated aliases array (we typically use urlAliases) with null included as its only item. Other versions have null as the only alias in their urlAliases array. Both cases are unexpected and probably shouldn't exist. This PR fixes what we expect from the database.


rayangler (approved):
Mongo CLI data loads successfully! I think we should just remove the bad data from the databases entirely/immediately, but this is a good intermediate step, just in case

backend/api/models/version.py
```diff
@@ -8,7 +8,7 @@ class Version(BaseModel):
     id: PydanticObjectId = Field(default_factory=PydanticObjectId)
     git_branch_name: str = Field(alias="gitBranchName")
     active: bool
-    url_aliases: list[str] | None = Field(alias="urlAliases")
+    url_aliases: list[str | None] | None = Field(alias="urlAliases")
```
rayangler, line 11: We also shouldn't keep/allow this in the long-term. A list of null values doesn't fit our intended schema and is kind of weird to have for our use case

PR #65, opened 2022-10-10
DOP-3294: save version without changes bug

mmeigs:
### Ticket

[DOP-3294](https://jira.mongodb.org/browse/DOP-3294)

### Notes

Whenever a version is edited and saved without actually changing any of the values of the fields, the operation errors out

Added a falsy check to ensure that then `existingVersions` variable always holds a valid array value.


carooobregon (approved):
yay for error handling! lgtm :)

PR #66, opened 2022-10-10
DOP-3300: Unable to create groups in repos with no existing groups

mmeigs:
### Ticket

[DOP-3300](https://jira.mongodb.org/browse/DOP-3300)

### Notes

Unable to create groups in repos with no existing groups.

Added check to see if groups property was null/falsy before pushing new group to repo.


backend/api/models/repo.py
```diff
@@ -41,6 +41,10 @@ class RepoVersionsView(BaseModel):
 
 async def insert_new_group(repo: Repo, group: Group) -> Repo:
     await repo.update({"$push": {"groups": group}})
```
rayangler, line 43: Might need to delete this line? We handle this in the `else` statement

rayangler: All good!

PR #67, opened 2022-10-11
DOP-3295 - Fix table styling

carooobregon:
**Ticket:**
[DOP-3295](https://jira.mongodb.org/browse/DOP-3295)

**Description:**
Tables have a small bug in which when the headings adjust dynamically based on the width of the current table's content for ex. when you drag and drop an element with a really long title or when there's just one element on the table. This leads to some UI jumpiness which is an unwanted user experience.
Solved it by fixing table layout however there are some risks to this solution as we're using global css styles to target elements that are under Leafygreen components and could affect other instances in the application.


mmeigs (commented):
It looks fantastic!!! My only issue is that it doesn't have a min-width, so it can all disappear as you get smaller. I think you can fix this with a min-width on one of the containers, such as the main.

frontend/src/styles/global.css
```diff
@@ -18,6 +18,14 @@ table [data-rbd-drag-handle-context-id] {
   cursor: pointer;
 }
 
+table td div{
```
rayangler, line 21: This rule is targeting ALL divs nested within a table cell. If we want to isolate it to only affect the first div parent that the cell uses, we could write it as `td > div` instead of `td div`. This might fix the positioning/centering of the icon in buttons on a group form page.

I'm not sure if doing this would cause any other side effects for the fixes you introduced, so it'd be good to double-check this

rayangler (commented):
Looks good right now. Just a couple of things to try

frontend/src/styles/global.css
carooobregon: awesome! that helped fix the button positioning and also minimizes the risk :)

frontend/src/styles/global.css
```diff
@@ -18,6 +18,14 @@ table [data-rbd-drag-handle-context-id] {
...
+  display: unset!important;
+}
+
+table td button{
+  display: block!important;
+}
```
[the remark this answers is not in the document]
carooobregon: applying to the specific Cell components doesn't seem to be helping and applying it to the general `Cell` seems to be affecting the position of many other things around!

carooobregon:
> It looks fantastic!!! My only issue is that it doesn't have a min-width, so it can all disappear as you get smaller. I think you can fix this with a min-width on one of the containers, such as the main.

good catch! i added the min-width to the table to make it a little more specific if that's alright

frontend/src/styles/global.css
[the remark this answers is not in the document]
rayangler: Hm, weird. It seems to be applying locally for me when added to the [Cell component](https://github.com/mongodb/docs-hapley/blob/7390c92ef617ce55474de8f80341c9411a453db7/frontend/src/components/Table/Cell.tsx#L15-L17)'s css:

```js
const cellStyling = css`
  vertical-align: middle;

  > div {
    display: unset !important;
  }
`;
```

If it's not working, I'm okay with ignoring. We already know this global css works at least

frontend/src/styles/global.css
```diff
@@ -18,6 +18,10 @@ table [data-rbd-drag-handle-context-id] {
   cursor: pointer;
 }
 
+td > div{
+  display: unset!important;
```
rayangler, line 22: (Super nit) Can we add a space between `div` and `{`, and a space between `unset` and `!important`? Helps keep this part of the code a little bit more tidy and conforms to the conventions of our other css rules

rayangler (commented):
One super nit thing but otherwise, the current code looks good to me.

Not sure what's causing this, but it seems like the icons for the "active" and "inactive" boolean values on the versions table are shrinking when the viewport grows smaller.
Screen Shot 2022-10-11 at 2 29 25 PM
Curious if this is happening for you locally as well or if you have any insight on this

mmeigs:
> > It looks fantastic!!! My only issue is that it doesn't have a min-width, so it can all disappear as you get smaller. I think you can fix this with a min-width on one of the containers, such as the main.
> 
> good catch! i added the min-width to the table to make it a little more specific if that's alright

Could you make it 600px? 300px still really cuts things off! 

Also, it seems this has affected the Group Form page. When you shrink enough, you can't scroll horizontally anymore!

frontend/src/styles/global.css
carooobregon: yup!

carooobregon:
@rayangler super interesting catch! i just fixed it with  adding `flex-shrink: 0;` to the icons
@mmeigs changed it to 600px! the Group Form page doesn't have scrolling on staging either so I think the changes done here didn't have to do with it? is it scrolling for you on staging?

frontend/src/styles/global.css
rayangler: Seems like this change wasn't part of the latest commits!

rayangler (approved):
LGTM! Thanks for fixing the shrinking icons. I'd wait to merge until it looks good to Matt as well

mmeigs:
> 

You're right, the scrolling isn't a problem! It does look like it's still 300 though!

carooobregon:
@mmeigs @rayangler thanks for the approvals! just updated the commits :)

mmeigs (approved):
LGTM!

PR #68, opened 2022-10-11
DOP-3273: GitHub API

mmeigs:
### Ticket

[DOP-3273](https://jira.mongodb.org/browse/DOP-3273)

### Notes

The user typing a github branch name themselves is prone to error. With the Github API, the user will now be able to select a branch name from a select dropdown to lower mistakes in creation of versions.

A/C:
- user can choose from unused github branch names when creating or editing a version on the version form page
- warning message given when a chosen name does not correspond with a current and existing github branch name 

Extra info:
- This relies upon an environment variable, a secret access token from our github bot.


carooobregon:
this is so cool and much needed! just a couple of things
1. when adding a new version, when there's no branch selected i'm getting the "This Github branch name no longer exists." by default
2. when there's a branch that doesn't exist anymore, i get the "This Github branch name no longer exists."  and then i click on "Edit" the banner disappears and I can still select it from the dropdown and save
great work!!

mmeigs:
> this is so cool and much needed! just a couple of things
> 
> 1. when adding a new version, when there's no branch selected i'm getting the "This Github branch name no longer exists." by default
> 2. when there's a branch that doesn't exist anymore, i get the "This Github branch name no longer exists."  and then i click on "Edit" the banner disappears and I can still select it from the dropdown and save
>    great work!!

Aha, that first one is something I didn't catch. Thank you!
The second... I kept the old (nonexistent) github branch name in the dropdown on purpose, so that the user would have the option to choose it again even if to keep it the same. This might be a conversation we all have about what the A/C is for this case. I'm torn about it. Luckily, once we change all the wrong names, this should be a very rare edge case.

carooobregon:
> > this is so cool and much needed! just a couple of things
> > 
> > 1. when adding a new version, when there's no branch selected i'm getting the "This Github branch name no longer exists." by default
> > 2. when there's a branch that doesn't exist anymore, i get the "This Github branch name no longer exists."  and then i click on "Edit" the banner disappears and I can still select it from the dropdown and save
> >    great work!!
> 
> @carooobregon Aha, that first one is something I didn't catch. Thank you! The second... I kept the old (nonexistent) github branch name in the dropdown on purpose, so that the user would have the option to choose it again even if to keep it the same. This might be a conversation we all have about what the A/C is for this case. I'm torn about it. Luckily, once we change all the wrong names, this should be a very rare edge case.

ah interesting yeah i see your point! and yeah as you said, once people use the branch names from the dropdown this will be a rare edge case

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -78,6 +82,20 @@ const aggregateUsedValues = (versions: Version[], activeVersion: Version | null)
...
+  allRepoBranches: string[],
+  existingVersions: Version[],
+  currentGitBranchName?: string
+) => {
+  const existingVersionsNames = existingVersions.map((version) => version.gitBranchName);
+  const remainingRepoBranchNameOptions = allRepoBranches.filter(
+    (repoBranchName) => !existingVersionsNames.includes(repoBranchName)
+  );
+
+  if (currentGitBranchName) return [currentGitBranchName, ...remainingRepoBranchNameOptions];
```
rayangler, line 95: Seeing a small bug where if I use the dropdown to click on a branch, and then open the dropdown again, the same git branch name is shown twice. Would probably be good to ensure uniqueness to avoid user confusion

rayangler (commented):
Looks really great! Tested it out locally and encountered a bug. The code looked good from what I saw, but I'll take another look tomorrow

frontend/package.json
```diff
@@ -52,6 +52,7 @@
     "@leafygreen-ui/typography": "^13.0.1",
     "gatsby": "^4.20.0",
     "gatsby-plugin-emotion": "^7.17.0",
+    "octokit": "^2.0.7",
```
rayangler, line 55: (No change needed, just for discussion) Curious if there's a reason why we use `octokit` here instead of [`@octokit/rest`](https://www.npmjs.com/package/@octokit/rest)? No strong preference for which we use at the moment, but it seems like `@octokit/rest` might be a lightweight version focused on data fetching

.drone.yml
```diff
@@ -62,6 +62,8 @@ steps:
     image: plugins/kaniko-ecr
     environment:
       GATSBY_API_URL: https://hapley.docs.staging.corp.mongodb.com/api/v1
+      GITHUB_ACCESS_TOKEN:
+        from_secret: github-access-token
```
rayangler, line 66: (Nit) Is it possible to make the secret snake case? All of the other secrets are currently in snake case and it might be good to stay consistent for the future. If not, it's okay too

environments/frontend-staging.yaml
```diff
@@ -1,6 +1,9 @@
 service:
   targetPort: 3000
 
+envSecrets:
+  GITHUB_ACCESS_TOKEN: github-access-token
```
rayangler, line 5: (Nit) Same thing about the snake case

rayangler (commented):
Left a comment for discussion based on one finding, and a few nits. As always, feel free to ignore if the nit isn't worth it.

I'm also unsure if the infrastructure changes work as expected now in this PR. Please feel free to ping me when ready!

frontend/package.json
mmeigs: huh, I hadn't even seen this! I was going through the Github API docs and all of its examples were using the larger octokit package. I'll try it out.

mmeigs:
> Left a comment for discussion based on one finding, and a few nits. As always, feel free to ignore if the nit isn't worth it.
> 
> I'm also unsure if the infrastructure changes work as expected now in this PR. Please feel free to ping me when ready!

^^ you're correct. I still haven't completely figured out how to access the secret yet in the staging environment.

mmeigs:
Well, multiple problems solved!! The environment variable is now active and working on the staging site, and because I have moved the logic to the backend, it will actually stay a secure secret. Thanks for holding out for me!

frontend/src/components/SideNav.tsx
```diff
@@ -57,15 +57,12 @@ const fullNavStyle = LeafyCSS`
 `;
 
 const SideNav: React.FC = () => {
-  let entitledRepos: string[] = useHapleyApi('/repos/')?.repos;
+  const { entitledRepos: repos } = useContext(ReposContext);
+  const entitledRepos = [...new Set((repos || []).map((repo) => repo.split('/')[1]))];
```
rayangler, line 61: Just curious what the reason for this change is? I assume it has something to do with the previous iteration of Github API on the frontend, but just wanted to check

frontend/tests/utils/mock-hapley-api.ts
```diff
@@ -1,7 +1,11 @@
+import { getAvailableBranches } from './mockData/github';
 import { getGroups } from './mockData/groups';
 import { GET_REPOS } from './mockData/repos';
 import { getVersions } from './mockData/versions';
 
+// Undesirable necessity for now to prevent infinite loop in testing for useEffect dependency equality
+const githubResult = getAvailableBranches('docs');
```
mmeigs, line 7: Here is the not-ideal solution! ^ I would have preferred to call getAvailableBranches(repoName) on line 26, but that oddly leads to the infinite loop within the VersionForm component in the useEffect that I'll also tag above. 
- I found that creating the result of the mock github call above a way to ensure that it is the same/equal value each time so that it doesn't cause a constant call of the useEffect that depends on it.

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -150,13 +171,36 @@ const VersionForm: React.FC<VersionFormProps> = ({
...
+  useEffect(() => {
+    if (!gitBranchNames || !gitBranchNames.length) return;
+    const availableGitBranchNames = getAvailableGitBranchNames(
+      gitBranchNames,
+      existingVersions,
+      initialVersion?.gitBranchName,
+      formState[GIT_BRANCH_NAME]
+    );
+    setGitBranchNameOptions(availableGitBranchNames);
+  }, [gitBranchNames, existingVersions, formState[GIT_BRANCH_NAME]]);
```
mmeigs, line 192: This is the useEffect dependency array that contains the result (gitBranchNames), and this fires over and over again without stop if I write the mock-hapley-api as I want it.

rayangler (commented):
Looks good! Just flagging some stuff I noticed (in particular: one potential frontend bug)

backend/api/v1/routes/github.py
```diff
@@ -0,0 +1,28 @@
...
+
+router = APIRouter()
+
+
+@router.get(
+    "",
+    response_model=list[str],
+    description="Get all github branches for a specific repo"
+)
+async def fetch_github_branches(repo: Repo = Depends(find_one_repo), entitled_repos=Depends(get_user_entitlements), settings: Settings = Depends(get_settings)):
```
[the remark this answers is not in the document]
mmeigs: Oh lovely! Thanks!

frontend/src/components/SideNav.tsx
mmeigs: You're right. For the previous iteration it was useful to have as an overarching context, but now it's unnecessary. I've removed the entire context! Thanks!

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -150,13 +171,36 @@ const VersionForm: React.FC<VersionFormProps> = ({
...
   }, [initialExistingVersions]);
 
+  const [gitBranchNameOptions, setGitBranchNameOptions] = useState<string[]>([]);
+  const gitBranchNames = useHapleyApi(`${versionsEndpoint}/github`);
+  useEffect(() => {
+    if (!gitBranchNames || !gitBranchNames.length) return;
+    const availableGitBranchNames = getAvailableGitBranchNames(
+      gitBranchNames,
+      existingVersions,
+      initialVersion?.gitBranchName,
```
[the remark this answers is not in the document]
mmeigs: Yes, that seems to work best!

rayangler (approved):
LGTM! Thanks so much for working on this. I like the API usage on the backend rather than the frontend

PR #69, opened 2022-10-13
DOP-3301: Drag and Drop API now using wholesale reordering

mmeigs:
### Ticket

[DOP-3301](https://jira.mongodb.org/browse/DOP-3301)

### Notes

Drag and Drop functionality has funky reordering bug. The leading hypothesis of the problem is the usage of indices to reorder the versions and groups rather than something less error-prone when multiple users are working.

First step is to change this API (add an API) to handle wholesale reordering. I've kept the old API as a PATCH method, and incorporated the wholesale reorder as the PUT method on these routes. 

Future steps: adding buttons or notifications/steps to the reorder itself. Will be handled by [DOP-3299](https://jira.mongodb.org/browse/DOP-3299)

More to talk about: I do believe there could possibly be further errors during high usage times (meaning UAT) when users' data won't be synced because of parallel changes. Although I think in use this will be very rare, we could look into either polling or implemented useSWR for the Repo page data.


backend/Pipfile
```diff
@@ -27,6 +27,6 @@ python_version = "3.10"
 
 [scripts]
 app = "uvicorn main:app --reload"
-test = "pipenv run pytest --cov=./api --cov-config=setup.cfg api/tests"
+test = "pipenv run pytest -s --cov=./api --cov-config=setup.cfg api/tests"
```
rayangler, line 30: Just curious what this `-s` option does!

backend/api/models/repo.py
```diff
@@ -1,3 +1,5 @@
+from typing import List
```
rayangler, line 1: We could use Python's native `list` datatype instead of `typing.List`. Or do we want to use `typing.List` for a specific reason?

rayangler (commented):
Overall looks good to me! Just one thing about a data type

rayangler:
> More to talk about: I do believe there could possibly be further errors during high usage times (meaning UAT) when users' data won't be synced because of parallel changes. Although I think in use this will be very rare, we could look into either polling or implemented useSWR for the Repo page data.

Agreed. We should definitely check to make sure that no changes were made to the data prior to accepting a user's changes.

backend/Pipfile
mmeigs: Ah yes, I decided to leave it in, but whatever you feel! It allows logs to the console to show during test runs! It's helped me, but we don't have to keep it

backend/api/models/repo.py
mmeigs: Gotcha!

mmeigs:
> > More to talk about: I do believe there could possibly be further errors during high usage times (meaning UAT) when users' data won't be synced because of parallel changes. Although I think in use this will be very rare, we could look into either polling or implemented useSWR for the Repo page data.
> 
> Agreed. We should definitely check to make sure that no changes were made to the data prior to accepting a user's changes.

Aha, interesting. Do you think this involves sending the old and new orders to the backend?

backend/Pipfile
rayangler: Oh, I'm fine with keeping it. I was just asking what it did because I wasn't sure, and looking it up didn't immediately help

rayangler:
> Aha, interesting. Do you think this involves sending the old and new orders to the backend?

We probably wouldn't need to send anything to the backend immediately? I was thinking more like:
1) User is on the `docs` repo page
2) User tries to reorder the versions and submits
3) Frontend fetches current data. If the data is the same, frontend submits data to the backend for successful update. If the data is different, the frontend prevents the submission and the user should be expected to refresh the page or something.

Not sure if this is optimal or ideal UX, but it was just an idea if we want to ensure that the data being updated is most up to date in the future, without completely overwriting someone else's changes. But like you said, this would probably be a rare occurrence. Something to discuss offline!

rayangler (approved):
LGTM!

PR #70, opened 2022-10-13
DOP-3187 - Error toast

carooobregon:
**Ticket**

[DOP-3187](https://jira.mongodb.org/browse/DOP-3187)

**Description:**
We want to reuse the Toast element to display errors to user when their request is not successful. Pretty much used a similar approach to the successful toast messages but changed some things for reusability.


frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -160,12 +160,18 @@ const GroupForm: React.FC<GroupFormProps> = ({
   }, [group, groups, versions]);
 
   // Wrapper for the toast message for group-related operations.
-  const triggerSuccessToast = (
-    callback: (itemType: string, itemName: string, repoName: string) => string,
-    itemName: string
+  const useToast = (
+    callback: (param1?: string, param2?: string, param3?: string) => string,
+    variant: Variant,
```
mmeigs, line 165: I think you can use ToastVariant and then we can get rid of the duplicate import too!

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -267,15 +272,16 @@ const GroupForm: React.FC<GroupFormProps> = ({
     if (modalType === 'delete') {
       // Deletion should only be possible if the group exists.
       if (isNew || !group) {
+        useToast(getStatusMessage.error, ToastVariant.Warning, undefined);
```
mmeigs, line 275: Since the third parameter is optional, you can get rid of the undefined!

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -172,12 +172,18 @@ const VersionForm: React.FC<VersionFormProps> = ({
   };
 
   // Wrapper for the toast message for version-related operations.
-  const triggerSuccessToast = (
-    callback: (itemType: string, itemName: string, repoName: string) => string,
-    itemName: string
+  const useToast = (
+    callback: (param1?: string, param2?: string, param3?: string) => string,
+    variant: Variant,
```
mmeigs, line 177: Same here with not needing the two imports!

mmeigs (commented):
Left some comments for now! I'm going to look closer later!

frontend/src/components/Forms/VersionForm.tsx
```diff
@@ -172,12 +172,18 @@ const VersionForm: React.FC<VersionFormProps> = ({
   };
 
   // Wrapper for the toast message for version-related operations.
-  const triggerSuccessToast = (
-    callback: (itemType: string, itemName: string, repoName: string) => string,
-    itemName: string
+  const useToast = (
+    callback: (param1?: string, param2?: string, param3?: string) => string,
```
rayangler, line 176: I'm a little wary of this callback function. It would be preferred to have more descriptive argument names and not have all of them need to be optional.

I'm wondering if it would be better to completely separate the success and error calls to the toast? The reason why the success messages are grouped together is because they all share the same requirements to construct their messages, but certain words are different. The error message seems different enough to warrant its own separate thing. Same for the `GroupForm`.

What do you think?

frontend/src/components/Forms/GroupForm.tsx
carooobregon: ohh nice for future reference!

frontend/src/components/Forms/VersionForm.tsx
carooobregon: i agree! changed it to separate things

frontend/src/components/Forms/GroupForm.tsx
```diff
@@ -168,6 +168,11 @@ const GroupForm: React.FC<GroupFormProps> = ({
     triggerToast(ToastVariant.Success, toastMessage);
   };
 
+  const triggerErrorToast = (message: string): void => {
+    const toastMessage = getErrorMessage(message);
+    triggerToast(ToastVariant.Warning, toastMessage, ERRORS.TOAST_TITLE);
```
mmeigs, line 173: I think you forgot to add the TOAST_TITLE constant back in ERRORS after all the git stuff!

carooobregon: thanks for the heads up!

mmeigs (approved):
LGTM!

PR #71, opened 2022-10-14
Json parse bug

mmeigs:
### Notes

Bad code that wasn't caught in my last PR. Quick fix back to how it was before. This was causing issues on creation and deletion of versions and groups.


carooobregon (approved):
lgtm!

PR #73, opened 2022-10-17
DOP-3299: Added reorder toast with Undo button, added error reorder toast

mmeigs:
### Ticket

[DOP-3299](https://jira.mongodb.org/browse/DOP-3299)

### Notes

After first bug bash, consensus is to add more control and user notification to the reorder drag and drop process on the Repo show page. After discussions with design, this PR adds a toast to let the user know the version/group was reordered and gives them an easy way to undo the change.

An error toast for a failed reordering was also added.


rayangler (approved):
LGTM!

I did noticed one small visual bug where a toast that isn't of the success variant will sometimes flicker from whatever variant it was (note, for example) to the success variant when it's about to close. I added a comment pinpointing where I think that's coming from and a suggestion to handle it. Happy to have this handled separately to avoid breaking anything now before the bug bash though since it's a pretty small thing!

PR #74, opened 2022-10-17
DOP-3317: Throw error on invalid user email

rayangler:
### Ticket

DOP-3317

### Notes

* Although the original ticket was about identifying a user through an email address regardless of whether it's mongodb or 10gen, it was confirmed by the Kanopy team that the JWT should always have a mongodb email address (unless there's something wrong with their system). As such, the logic to determine a user by email will remain the same.
* It was discovered that no error message was being sent back to the frontend despite not having found a matching user in the `entitlements` collection. Turns out we didn't check for a user's existence and we'd just return a `None` object as a response. This PR introduces a fix for this.
* To test locally, try changing your email address in your local entitlements collection. The new `UserNotFound` error should be returned.


rayangler (commented):
Notes/annotations

mmeigs (approved):
Looks wonderful, thanks!

PR #75, opened 2022-10-18
DOP-3316: Reorder Undo bug, Toast Style bug

mmeigs:
### Ticket

[DOP-3316](https://jira.mongodb.org/browse/DOP-3316)

### Notes

This PR fixes two known bugs.

- The user should be able to persist changes when undoing a reorder operation of the versions and groups on the repo show page
- The toast style should not flash between styles when opening or closing.


frontend/src/context/toast-context.tsx
```diff
@@ -37,16 +37,18 @@ const ToastContext = createContext<ToastContextValue>({
...
     (variant: Variant, body: string | ReactElement, title?: string) => {
       setToastProps({ variant, body, title });
+      setOpen(true);
     },
     []
   );
 
   const closeToast = useCallback(() => {
+    setOpen(false);
     setToastProps(null);
```
rayangler, line 52: I'm still seeing the toast change variant for a quick moment on close. Removing `setToastProps` here should probably resolve this, since it avoids triggering the default "success" variant

frontend/src/components/Repo.tsx
```diff
@@ -95,49 +95,33 @@ const Repo: React.FC<{ repoName: string }> = ({ repoName }) => {
...
-            resetOrder(itemType, lastState);
-            closeToast();
+          onClick={async () => {
+            const lastState = retrieveLastState();
+            resetReorder(itemType, lastState);
+            const response = await repoItemAPI('PUT', `/repos/${repoName}/${itemType}s`, () => {}, {
+              newOrder: lastState
+            });
+
+            // Reset to previous state if API request errors
```
rayangler, line 119: This is a really cool edge case! I like this.

Just wondering: did you encounter this locally and/or is there a particular example of when this would happen? Or is this more of a "just in case" ?

rayangler (commented):
Really great job on this!

frontend/src/components/Repo.tsx
mmeigs: Never seen it happen! (unless I force it) So, a just in case! I mean, the server would have to fail pretty quickly or something, but I thought I should always have an error handler of some kind!

frontend/src/context/toast-context.tsx
mmeigs: hahaha noooooo! There are a ton of weird edge cases with this! Could you actually let me know how to reproduce that flicker?

mmeigs: Nevermind! Saw it

mmeigs: Ok, you're right! I must have gotten rid of the other error with another piece of code. Thanks for seeing this!

rayangler: Ah, was about to send a video to reproduce. But yeah, it's a super small and quick thing that happens... that ideally shouldn't happen. The timing of state changes can be a little silly

frontend/src/components/Repo.tsx
rayangler: Makes sense! Thanks!

PR #76, opened 2022-10-18
DOP-3315: Alphabetize the entitled repos list

mmeigs:
### Ticket

DOP-3315

### Notes

Based off feedback from Bug Bash II, it was decided that for ease of use it would be best to have the repos in the sidenav be alphabetized. 

This work was completed on the backend to keep that work more efficient and closer to the state of truth. Can be switched easily if we feel it's more of a cosmetic change and therefore better in the frontend component.


rayangler (approved):
Very cool

PR #80, opened 2022-10-21
DOP-3329: Defaults made more apparent

mmeigs:
### Ticket

DOP-3329

### Notes

From UAT feedback, we were asked to make defaults more apparent to the user. ie the Selector Label and the URL Slug are defaulted to the Github Branch Name if they are not filled in.

In discussions with design, it has been decided to try to use a dynamic placeholder value for these two fields that will match the Github Branch Name. This will be less text to take in, but hopefully understandable to the user.

Extra: also after speaking with design, this PR addresses the tense of a Version form property that seemed incorrect. "Published Original Branch Name" has now been changed to "Publish Original Branch Name."


rayangler (approved):
Very cool

PR #81, opened 2022-10-26
DOP-3333: DnD efficiencies

mmeigs:
### Ticket

[DOP-3333](https://jira.mongodb.org/browse/DOP-3333)

### Notes

After spending a lot of time scouring the React Profiler results of our drag and drop, I came up with three different changes to bring re-renders down. 
* ensuring, through a more specifically-designed prop, that only the cells in the row being dragged had their height and width frozen. This keeps the other cells on the table from being re-rendered.
* removing tooltips from the Menu icon drag and drop handle which said "Reorder the {version | group}" - this oddly was time-consuming
* Changing a couple LG components to simple html components. The Cell and Body seemed to also have less efficient re-rendering than their simple counterparts.


frontend/src/components/Table/Cell.tsx
```diff
@@ -40,9 +40,9 @@ const Cell = ({ children, isDragging }: CellProps) => {
   }, [isDragging]);
 
   return (
-    <LGCell ref={cellRef} className={cx(cellStyling)}>
```
rayangler, line 43: Lint: we can remove `LGCell` as an import

frontend/src/components/Table/styles.tsx
```diff
@@ -11,7 +11,7 @@ export const cellStyle = LeafyCSS`
   text-overflow: ellipsis;
 `;
 
-export const TableCell = styled(Body)`
```
rayangler, line 14: Lint: we can remove `Body` as an import

rayangler (approved):
LGTM! Just a couple of linting-related stuff

PR #83, opened 2022-10-27
Select option-order

mmeigs:
### Notes

There is no ticket for this, but it has been bothering me for a day or two, and I should have included it in the Github PR.

Currently the selected gitBranchName and the "initial" gitBranchName are at the bottom of the list of options in the dropdown.

This PR simply makes sure that these "used" options are at the top of the dropdown options.


rayangler (approved):
Nice, I like this

PR #86, opened 2022-10-28
Make reordering list payload a generic model

rayangler:
### Context

IDs were being saved into the database as strings instead of as `ObjectIds` due to reordering. This was most likely occurring due to the lack of typing when it came to the payload class that housed the new order of versions/groups.

### Solution

The payload class was converted to be generic to allow for typing. This helps ensure that everything in the payload list of items to reorder are specific to the data type we want (`Version` or `Group`)


backend/api/dependencies.py
```diff
@@ -48,7 +48,7 @@ async def get_user_entitlements(
     ).project(PersonalRepos)
     if not entitled_repos:
         raise UserNotFound()
-    entitled_repos.repos.sort(key=lambda x: x.split('/')[1])
+    entitled_repos.repos.sort(key=lambda x: x.split("/")[1])
```
rayangler, line 51: Hm, not sure what happened here. Seems like my local linter/formatter did this... I have to look into this for myself

mmeigs (approved):
LGTM! Thank you!!!

PR #87, opened 2022-10-31
Update backend readme for local development

rayangler:
### Context

General updates to the backend's `README`. The order of certain sections would sometimes cause hiccups during onboarding. This PR hopes to refine the instructions to make it more streamlined for future collaborators.

### Notes
* [Link to the formatted README](https://github.com/mongodb/docs-hapley/blob/62e56467e314f9be2b3d547180a4f4c18ac30a4c/backend/README.md), for convenience.
* Adds a small section about environment variables. It mostly highlights `.env.example`, which is a useful file that we have but don't even mention!
* Removes documentation about the shared `hapley-dev` database. It should be preferable for us to connect to our local databases so that WIP changes don't cause any hiccups for others.
* Moved and updated the sample JWT section to be below the database section. If I remember correctly, the backend would error due to missing database envs when trying to generate the JWT, so it makes sense to reorder to minimize this from happening during initial setup.


mmeigs (approved):
LGTM! Thanks!

